PR #6, opened 2015-04-01
Use fewer macros without losing performance

reem:
More types, fewer macros! No performance loss!

If you're wondering about the flurry of PRs it's because I'm about
to fork this code to work with @cgaebel's iobuf library, so I'm
just making some changes before I get started.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,8 +2,7 @@
 #![feature(core)]
 #![allow(unused_assignments)]
 
-use std::str;
-
+use std::{str, mem, raw};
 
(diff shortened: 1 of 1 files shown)


reem:
err, wait, don't merge this, I made this PR to hastily

reem:
@seanmonstar should be ready to go now

seanmonstar:
Tests errored.

Curious, why inline functions over macros?

reem:
Typo, fixed it.

Functions are clearer than macros imo when you don't need macro features like return from caller.

PR #11, opened 2015-08-15
skip empty lines before Request-Line

bonifaido:
According to the [specification](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.1) they might come. I have observed this when a hyper server started to give "request error = Version" messages.


seanmonstar:
Thanks for the pull request! The specification also says newlines can be either CRLF or just LF, so I've adjusted it to allow both. I merged that into master here: https://github.com/seanmonstar/httparse/commit/5ad25593e76baf399622d77b2df66720fa76e33e

PR #13, opened 2015-12-23
Stop parsing requests with invalid method or path delimiters

ebarnard:
Previously requests of the form `GET\n/ HTTP/1.1\r\nHost: foo.bar\r\n\r\n` would parse.


seanmonstar:
Thank you! Staring at this PR, I can't recall what I was possibly thinking...

PR #17, opened 2016-08-04
Don't use std

jethrogb:
Thanks to the no-copy design, this crate has no dependencies on things in `std`, and it's trivial to make it work with `no_std` configuration.


seanmonstar:
Good point!

PR #21, opened 2016-09-12
Implement Display for Error and InvalidChunkSize

17dec:
So that the errors can be passed to any macro/function that expects the Display trait.

Ideally these error types would also implement the Error trait, but that would require the std crate.


seanmonstar:
Thanks!

PR #22, opened 2016-10-17
Add no_std feature, impl std::error::Error

abonander:


abonander:
@seanmonstar Are you waiting to publish a new version or did you forget again? ;P

seanmonstar:
Something tickles me weird about having the `no_std` be a feature. I wanted to think about it some to see if there's something better.

abonander:
It's a common convention, I made it a default so it didn't break any existing usage expecting it to work with `core` right out of the box.

abonander:
@seanmonstar Would you prefer to leave the `#![no_std]` unconditional and just have a `std` feature? I think that would work:

    #![no_std] 

    #[cfg(feature = "std")]
    extern crate std;

    #[cfg(feature = "std")]
    impl ::std::error::Error for Error { /* ... */ }

seanmonstar:
Ignore me, it's perfect. Thanks!

jethrogb:
The convention is to have a feature named `std`, see [this comment](https://github.com/rust-lang-nursery/rand/pull/108#issuecomment-231556510) for rationale.

seanmonstar:
@jethrogb so, have this instead:

```toml
[features]
default = ["std"]
std = []
```

And anyone wanting to use it in a `no_std` environment sets `default-features = false`.

seanmonstar:
If that seems correct, I've made a PR to do that, and I can release the new version today https://github.com/seanmonstar/httparse/pull/23

PR #23, opened 2016-11-07
convert no_std to std feature

seanmonstar:


src/lib.rs
```diff
@@ -14,14 +14,10 @@
 //! 1.6 times slower than pico. Improvements can be made as a `likely`
 //! intrinsic, and simd, are stabilized in rustc.
 
-#[cfg(test)] extern crate core;
+#[cfg(feature = "std")] extern crate std as core;
```
jethrogb: You can just do `extern crate core`

jethrogb:
Looks good to me!

Issue #15, opened 2016-02-13
Remove trailing whitespace from header field values


pyfisch:
[RFC7230#Field parsing](https://tools.ietf.org/html/rfc7230#section-3.2.4) states that parsers should remove leading in trailing whitespace from header field values. Currently httparse only removes leading whitespace, removing trailing whitespace would make httparse easier to use for other crates.

The following two requests should be parsed the same:
- `GET / HTTP/1.1\r\nHost: foo.com\r\nUser-Agent: foobarsoft\r\n\r\n`
- `GET / HTTP/1.1\r\nHost: foo.com\r\nUser-Agent: foobarsoft \t \t \r\n\r\n`

seanmonstar:
Something to keep in mind if implementing this is to do so without losing
much performance.

seanmonstar:
This was done in #48.

PR #26, opened 2016-12-03
Trailing whitespace

lpbak:
Proposed fix for issue #15.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -507,6 +507,7 @@ fn parse_headers_iter<'a, 'b>(headers: &mut &mut [Header<'a>], bytes: &'b mut By
             }
 
             let mut b;
+            let mut trailing_whitespace : usize = 0;
 
             'value: loop {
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Neat! Do you know what the difference is with the benchmark?

lpbak:
I've run this change against the benchmark (after dealing with #27) and unfortunately, the impact seems to be significant. I'll try to look into it further.

lpbak:
I've been able to decrease benchmark times somewhat, but the impact of this change is still visible. Please, decide whether this impact is acceptable, and act on this PR accordingly.

seanmonstar:
I'm sorry, not ignoring, I'll look into the benchmarks in the next couple days.

seanmonstar:
I just tried this out, and on my machine, I see almost a 157% increase in parse time (638ns vs 1006ns), which is probably too much.  Additionally, I expect client and servers to selectively only trim header values that need to be trimmed. Considering that trimming all header values is not necessarily something that needs to be done, it feels too expensive.

If we could keep the performance cost much much lower, it may be worth it. Thanks @lpbak for trying this out!

PR #31, opened 2017-01-10
Added header function to extract specific value

hanscj1:
Added a function that supports UniCase and extracts the specific value for a given header.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,9 @@ license = "MIT/Apache-2.0"
 description = "A tiny, safe, speedy, zero-copy HTTP/1.x parser."
 repository = "https://github.com/seanmonstar/httparse"
 
+[dependencies]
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks for the pull request!

Is there a reason to include this in this crate, instead of just implementing this where one needs it? An issue I see here is that the unicase crate needs std, and httparse tries to work without it.

hanscj1:
I'll take a look at doing that in my library. It was just easier at the time. thx

hanscj1:
Just curious, I have seen several libraries lately working toward !std, why? thx

seanmonstar:
I don't have any special interest in bending over backwards to make it work with `no_std`, but httparse just didn't need anything in `std`. The benefit is that people can use those libraries on small hardware, where executable size may be important, or access to a heap may not even exist.

PR #32, opened 2017-01-27
Change type of name to slice

GuillaumeGomez:
Is the `Str` struct necessary or not? This is the main question in here.


seanmonstar:
I don't quite understand why make this change.

Issue #42, opened 2018-06-22
Build broken on 1.27.0 with target `i686-linux-android`


thomcc:
The specific error I get is "error[E0425]: cannot find function `_tzcnt_u64` in this scope". I assume this is because it's a 32 bit platform but it could be something else.

seanmonstar:
Welp, yep, that's busted. The AVX2 function is configured to exist on both `x86` and `x86_64`, but `_tzcnt_u64` doesn't exist in `std::arch::x86`!

thomcc:
Is that a bug in the stdlib? I think both should just be the `tzcnt` instruction, if you support one you should support the other.

seanmonstar:
I don't know if it's a bug in libstd or not... There does appear to be `std::arc::x86::_tzcnt_u32`, which I assume is the same operation, and both just actually use a word?

thomcc:
The actual instruction is overloaded on what you pass it (sort of, it will be encoded differently). See https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=tzcnt&expand=5536,5535 where the instruction is `tzcnt r64, r64` vs `tzcnt r32, r32`.

I don't think any chips exist that are actually 32 bit and support these instructions though, so IDK how it would work on them.

seanmonstar:
Instead of hastily swapping to `_tzcnt_u32` or something without considering the consequences, I've pushed #43 that just disables AVX2 usage on x86. Is it possible to adjust your `Cargo.toml` to point at that to see if it fixes it?

thomcc:
Yep, that fixes it.

seanmonstar:
Alright, published v1.3.1! Thanks for reporting this.

iximeow:
judging by the `0xffffffff_00000000` mask [used with _tzcnt_u64](https://github.com/seanmonstar/httparse/blob/7920414/src/simd/avx2.rs#L108) i _think_ it could be removed and `_tzcnt_u32` used on both architectures. there's a corner case for processors that do not support BMI1 where the same instruction bytes instead mean `BSF` (with slightly different behavior on a 0 operand)

i'll try this out, see if tests pass, kick the tires etc and put up a PR for that. AVX2 can be used on x86 targets by using a 32-bit OS on modern processors, for example, so this is plausibly (if unlikely) reachable.

PR #43, opened 2018-06-22
disable avx2 on 32bit archs temporarily

seanmonstar:
As reported in #42, the `_tzcnt_u64` function isn't defined on `x86`, and so causes compilation errors. As a quick fix, this simply disables avx2 on that arch.


seanmonstar:
The CI failure is due to Rust nightly being broken, all the other jobs pass, and this fixes the issue.

PR #48, opened 2018-09-22
Exclude trailing whitespace in header value.

hannesg:
According to [rfc7230#section-3.2.4](https://tools.ietf.org/html/rfc7230#section-3.2.4):
> The field value does not include any leading or trailing whitespace:
> OWS occurring before the first non-whitespace octet of the field value
> or after the last non-whitespace octet of the field value ought to be excluded by
> parsers when extracting the field value from a header field.

Leading whitespace already gets removed in a loop before parsing the value.

See also https://github.com/hyperium/http/issues/245


seanmonstar:
Thanks!

PR #49, opened 2018-09-24
Add support for rare symbols

theikkila:
Added symbols are tested against nginx and they seem to work. (so they don't return 400)
WhatWG specification seems to be even broader than this:
https://url.spec.whatwg.org/#query-state


seanmonstar:
Thanks! :D

PR #50, opened 2018-09-24
Rare symbols generated test suite

theikkila:
Adds additional testsuite to #49  based on whatwg tests


seanmonstar:
Wow, thanks! I'm going to check this out locally and move these into a separate file :D

seanmonstar:
Merged manually in e30e85f0be73ffcdf383118d57c124d8b6aa793d, thanks again!

PR #64, opened 2019-10-31
provide example of parsing headers from a `Read`

euclio:
I found it really tricky to figure out how to do this, so I think that it would be nice to provide an example in the docs.


seanmonstar:
Thanks for trying to fix up the docs!

I'm personally not a fan of `BufRead`, and along with the likely case where async IO would be a good idea for HTTP, I'm inclined to *not* include such an example.

PR #75, opened 2020-07-03
Fix off-by-one error

Hawk777:
If len = pos + 8, then there are exactly eight bytes available.


seanmonstar:
Good catch! What do you think if we added a couple unit tests to this file that would catch this?

Hawk777:
Something like 7a7d9af6adab0983c335129f52066fea0499eea6?

PR #76, opened 2020-07-05
Correct `&` to `&&`

spinda:
`&` applied to two booleans is the non-short-circuiting logical AND. That works here, but `&&` is probably what was intended.


shirshak55:
i guess it would be better if we can write test so there wouldn't be such bug?

spinda:
Not possible as the behavior of the code is the same either way (`&` just doesn't apply short-circuiting).

PR #77, opened 2020-07-05
Add missing Eq derivations

Hawk777:
The derivation on Header is correct because comparisons of string slices and byte slices meet the requirements. The derivations on Request and Response are correct because they contain only string slices, byte slices, slices of Header objects (which are now Eq), primitives, and options (which are Eq if their contents are). Finally the derivation on Status is correct because #[derive(Eq)] on a generic only actually implements Eq if all generic parameters are Eq, which for Status is sufficient.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -212,7 +212,7 @@ pub type Result<T> = result::Result<Status<T>, Error>;
 /// `Complete` is used when the buffer contained the complete value.
 /// `Partial` is used when parsing did not reach the end of the expected value,
 /// but no invalid data was found.
-#[derive(Copy, Clone, PartialEq, Debug)]
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
 pub enum Status<T> {
     /// The completed result.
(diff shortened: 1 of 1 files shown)


seanmonstar:
I agree they're correct, I just have one question: why would we want these implementations? What case does this help solve?

Hawk777:
Given that `Eq` is an empty trait, AFAICT there is no downside to adding it. [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html) recommends eagerly implementing `Eq` where it is appropriate. It means that other people will then be able to `#[derive(Eq)]` on their own types that contain e.g. a `Request` object, which would not otherwise work; without this, those other people would have to `impl Eq for MyType {}` explicitly, which is significantly more verbose and also a little bit error-prone given that it leaves open the possibility that `Request` might stop being effectively `Eq` sometime and the other person not notice.

seanmonstar:
Yea, thats true. Thanks, I'll merge!

PR #78, opened 2020-08-10
 Replace deprecated 'try!' with '?'

eaufavor:
https://doc.rust-lang.org/std/macro.try.html

>  Deprecated since 1.39.0: use the ? operator instead


seanmonstar:
We can't make this change, because then it stops building on Rust versions before `?` became stable.

PR #80, opened 2020-10-10
Basic fuzzing using cargo-fuzz

ArniDagur:
Reading https://lobste.rs/s/bfjdfd/rust_curl_with_hyper#c_r8wiua made me curious to do some basic fuzzing on this crate using `cargo-fuzz`. None of the included targets found anything interesting though, which is good. Perhaps we should try to target the SIMD code directly, by exposing it publicly with a `fuzzing` feature or something like that?

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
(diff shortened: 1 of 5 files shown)


seanmonstar:
Thanks for adding these! It'd also be great to get some more specific testing of the SIMD parsers. I believe they can be forcefully enabled or disabled with environment variables in the build script.

PR #84, opened 2021-01-30
Allow obs-text characters in response status reason phrase

acfoltzer:
In particular, this avoids parsing failures when servers return accented ISO 8859-1 characters.

The existing comment noting the non-compliance had two suggestions for future improvement: return a `&[u8]`, or add specialized helpers such as `display()`. I decided to go with the former since the reason phrase appears largely unused in downstream dependencies like `hyper`.

This is unfortunately a breaking change due to the new type, but it's a very important change for interoperability with traffic in the wild.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -351,6 +351,8 @@ pub struct Response<'headers, 'buf: 'headers> {
     /// The response code, such as `200`.
     pub code: Option<u16>,
     /// The response reason-phrase, such as `OK`.
+    ///
+    /// Contains an empty string if the reason-phrase was missing or contained invalid characters.
     pub reason: Option<&'buf str>,
     /// The response headers.
     pub headers: &'headers mut [Header<'buf>]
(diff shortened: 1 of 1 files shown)


seanmonstar:
Do you know of specific websites that do this? (Maybe you can't share, that's fine...)

I'm bummed about the breaking change... I've only thought about this for a minute, but another option is to add some `reason_bytes: Option<&[u8]>` field...

acfoltzer:
> Do you know of specific websites that do this? (Maybe you can't share, that's fine...)

Yes, and maybe soon but not now :crossed_fingers: 

> I'm bummed about the breaking change...

Yeah, agreed. The only way I've thought up to avoid the breaking change is to return a static `""` if non-utf8 bytes are found, after still consuming the reason phrase from the bytes buffer. 

> another option is to add some `reason_bytes: Option<&[u8]>` field...

This would also be breaking because all the fields of `Response` are `pub` :disappointed:

seanmonstar:
> This would also be breaking because all the fields of Response are pub

Eek, you're right! Ugh, I wrote this crate so long ago, before I realized API traps were a thing...

> return a static "" if non-utf8 bytes are found, after still consuming the reason phrase from the bytes buffer.

This doesn't sound so bad, actually... Sure, ideally we'd just have them be opaque bytes, but, but! Maybe the bytes aren't really that useful anyways, and so we can just skip them...?

acfoltzer:
> This doesn't sound so bad, actually... Sure, ideally we'd just have them be opaque bytes, but, but! Maybe the bytes aren't really that useful anyways, and so we can just skip them...?

That'd be fine with us, after all we're only using this crate via `hyper` which doesn't use the reason phrase. Shall I update the PR with the empty string behavior?

seanmonstar:
Yea, that sounds good to me, thanks!

acfoltzer:
Pushed the new behavior in 1ca756c

src/lib.rs
```diff
@@ -477,28 +479,38 @@ fn parse_version(bytes: &mut Bytes) -> Result<u8> {
...
+                if !seen_obs_text {
+                    // all bytes up till `i` must have been HTAB / SP / VCHAR
+                    str::from_utf8_unchecked(bytes)
+                } else {
+                    // obs-text characters were found, so return the fallback empty string
+                    ""
+                }
             }));
-        } else if !((b >= 0x20 && b <= 0x7E) || b == b'\t') {
+        } else if !(b == 0x09 || b == b' ' || (b >= 0x21 && b <= 0x7E) || b >= 0x80) {
```
acfoltzer, line 510: I rearranged this conditional to line up better with the excerpt from the RFC, but can back it out if you'd like to avoid the churn

seanmonstar (approved):

PR #86, opened 2021-03-13
Reduce Usage of Unsafe

CalebLBaker:
Remove unnecessary usage of `unsafe` and replace some unsafe code with equivalent safe code.

The safe implementation of `iter::Bytes::slice_skip` proposed by this pull request has been benchmarked to be (slightly) slower than the currently used implementation, so if performance is an issue I can revert the changes to that particular method while still keeping my other changes.

`shrink` cannot be written without using unsafe (there are issues related to temporarily mutably borrowing the slice multiple times simultaneously), so I am ignoring it in this pull request. I might come back later and look at how `shrink` is used to see if something safer can be used instead.

I also ignored usage of `str::from_utf8_unchecked` since it seems apparent that whenever it is used in this crate the bytes passed in have already been checked for utf-8 compliance and so checking again would be redundant.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -50,7 +50,8 @@ impl<'a> Bytes<'a> {
     pub fn slice_skip(&mut self, skip: usize) -> &'a [u8] {
         debug_assert!(self.pos >= skip);
         let head_pos = self.pos - skip;
-        let (head, tail) = self.slice.split_at(head_pos);
+        let head = &self.slice[..head_pos];
+        let tail = &self.slice[self.pos..];
         self.pos = 0;
         self.slice = tail;
         head
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -1,5 +1,3 @@
-use core::slice;
-
 pub struct Bytes<'a> {
     slice: &'a [u8],
     pos: usize
@@ -25,14 +23,14 @@ impl<'a> Bytes<'a> {
     }
 
     #[inline]
-    pub unsafe fn bump(&mut self) {
+    pub fn bump(&mut self) {
         debug_assert!(self.pos + 1 <= self.slice.len(), "overflow");
         self.pos += 1;
     }
 
     #[allow(unused)]
     #[inline]
-    pub unsafe fn advance(&mut self, n: usize) {
+    pub fn advance(&mut self, n: usize) {
         debug_assert!(self.pos + n <= self.slice.len(), "overflow");
         self.pos += n;
     }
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks so much for filing this! I'll take a look through now :)

seanmonstar:
I was curious how the extra bounds checks would perform, so when I run the benchmarks of master vs this PR, here's what I see:

- **master**:
  
  ```
  test bench_httparse       ... bench:         569 ns/iter (+/- 103) = 1265 MB/s
  test bench_httparse_short ... bench:          67 ns/iter (+/- 1) = 1014 MB/s
  ```

- **PR**:
  
  ```
  test bench_httparse       ... bench:         701 ns/iter (+/- 67) = 1027 MB/s
  test bench_httparse_short ... bench:          80 ns/iter (+/- 1) = 850 MB/s
  ```

CalebLBaker:
That looks similar to the results that I got from running the benchmarks. I've played around with benchmarking different subsets of the pull request and the greatest portion of the performance difference definitely seems to come from slice_skip.

src/iter.rs
```diff
@@ -25,14 +23,14 @@ impl<'a> Bytes<'a> {
     }
 
     #[inline]
-    pub unsafe fn bump(&mut self) {
```
hkBst, line 28: This function can violate a struct invariant, so unless the `debug_assert` is changed to `assert` it must be marked as `unsafe` or it would be unsound.

src/iter.rs
```diff
@@ -25,14 +23,14 @@ impl<'a> Bytes<'a> {
...
+    pub fn bump(&mut self) {
         debug_assert!(self.pos + 1 <= self.slice.len(), "overflow");
         self.pos += 1;
     }
 
     #[allow(unused)]
     #[inline]
-    pub unsafe fn advance(&mut self, n: usize) {
+    pub fn advance(&mut self, n: usize) {
         debug_assert!(self.pos + n <= self.slice.len(), "overflow");
```
hkBst, line 34: Same as above, this function is `unsound` without `unsafe`.

src/iter.rs
```diff
@@ -45,18 +43,15 @@ impl<'a> Bytes<'a> {
...
         // not moving position at all, so it's safe
-        unsafe {
-            self.slice_skip(0)
-        }
+        self.slice_skip(0)
     }
 
     #[inline]
-    pub unsafe fn slice_skip(&mut self, skip: usize) -> &'a [u8] {
+    pub fn slice_skip(&mut self, skip: usize) -> &'a [u8] {
```
hkBst, line 50: Same as above, this function is unsound without `unsafe`.

seanmonstar:
A lot has been refactored since, so I'll close.

PR #91, opened 2021-03-22
Allow spaces between header name and colon (fixes #88)

### LINKED ISSUE #88: Allow spaces between header names and colons
Sorry, I edited my issue, mixed up two problems at once.

<hr>

```
curl -v "https://videoroll.net/vpaut_option_get.php?pl_id=6577"
```

Note the space after `Access-Control-Allow-Credentials`.

```
< HTTP/1.1 200 OK
< Server: nginx/1.16.0
< Date: Tue, 16 Mar 2021 09:03:39 GMT
< Content-Type: text/json;charset=utf-8
< Transfer-Encoding: chunked
< Connection: keep-alive
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials : true
< Expires: Tue, 23 Mar 2021 09:03:39 GMT
< Cache-Control: max-age=604800
```

What does [the spec](https://tools.ietf.org/html/rfc7230#section-3.2.4) says about this?

> No whitespace is allowed between the header field-name and colon.  In the past, differences in the handling of such whitespace have led to security vulnerabilities in request routing and response handling.  A server MUST reject any received request message that contains whitespace between a header field-name and colon with a response code of 400 (Bad Request).  A proxy MUST remove any such whitespace from a response message before forwarding the message downstream.

So first, I want to point out that the security vulnerabilities can't really happen through httparse (or at the very least, through Hyper), given Hyper does not keep around the literal text of the HTTP request, and thus those pesky spaces cannot be passed to anyone downstream AFAIK.

Second, and this is what matters to me (wearing my work hat), is that the spec itself implies that a proxy has to be able to parse those anyway to remove the headers, thus it needs to successfully be able to parse such a response. httparse (and thus Hyper) does not let any of that pass through, which is a problem for people implementing proxies.

Making a patch that makes those spaces not fail the entire parse is pretty trivial, but I wonder if we want a `relaxed_response_parsing` option of some sort. Such an option [exists](https://github.com/squid-cache/squid/blob/c64d8e60a3f04951509929c03ab57d99b2c068e7/src/HttpHeader.cc#L1461-L1491) in Squid.

<hr>

In the browser space, Firefox (and AFAIK Chrome too) happily ignore spaces between the header name and the colon, and will also ignore spaces in header names themselves (ignoring the whole "name: value" pair and just skipping to the next header in the response.

```
curl -v https://crlog-crcn.adobe.com/crcn/PvbPreference -X POST 
```

Note el famoso `Updated Preferences: []` header in the response.

```
> POST /crcn/PvbPreference HTTP/1.1
> Host: crlog-crcn.adobe.com
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200
< Updated Preferences: []
< Content-Length: 2
< Date: Tue, 16 Mar 2021 08:43:38 GMT
<
```

We can discuss whether or not to let that through at a later date.

nox:


nox:
Maybe we can put the whitespace skipping in its own `#[cold] #[inline(never)]` function to keep it off the happy path.

nox:
I have restored 1.10 compat.

src/lib.rs
```diff
@@ -250,6 +250,38 @@ impl<T> Status<T> {
...
+
+impl ParserConfig {
+    /// Sets whether spaces should be allowed after header name.
+    pub fn allow_spaces_after_header_name(&mut self, value: bool) -> &mut Self {
+        self.allow_spaces_after_header_name = value;
+        self
+    }
+
+    /// Parses a request with the given config.
+    pub fn parse_request<'headers, 'buf>(
```
seanmonstar, line 267: I suppose in a general sense, it makes sense to be able to pass some config when parsing requests. Just to be sure, in this specific case, there's no value in allowing spaces before colons in *requests*, since that's illegal and servers should reject it, right?

nox: Yeah I don't think I would use it in our proxy.

nox: I removed the config for parsing requests, we can always introduce that when we actually need it.

seanmonstar (approved):

PR #96, opened 2021-05-08
Fix parsing responses with no CRs

sfackler:
The status code would previously be folded into the first header name.

### DIFF
--- a/tests/uri.rs
+++ b/tests/uri.rs
@@ -24,26 +24,6 @@ macro_rules! req {
     )
 }
 
-macro_rules! res {
-    ($name:ident, $buf:expr, |$arg:ident| $body:expr) => {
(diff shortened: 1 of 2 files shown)


tests/uri.rs
```diff
@@ -24,6 +24,26 @@ macro_rules! req {
     )
 }
 
+macro_rules! res {
```
seanmonstar, line 27: Could probably be after these tests in `lib.rs`: https://github.com/seanmonstar/httparse/blob/5c385a9b1751f0734db24af731d1926e1d2bc731/src/lib.rs#L1118-L1136

I think these "uri" tests were ported from some parse test suite in Firefox or something...

seanmonstar (approved):

Issue #95, opened 2021-05-04
Need for `Request` to take a slice of `MaybeUninit<Header>`


Soveu:
Context: `hyper` passes [uninitialized array](https://github.com/hyperium/hyper/blob/master/src/proto/h1/role.rs#L131) of `httparse::Header` to `Request::new`.
This is undefined behavior and I discovered it while working on [Replacing `mem::uninitialized` with `MaybeUninit`](https://github.com/hyperium/hyper/issues/2532)

I acknowledge that changing `headers` field will be a breaking change, because it is public, so
is there a chance that a `Request` type copy could be implemented, but with `headers: &mut [MaybeUninit<Header>]`?

EDIT: I just realized that author of `httparse` is also author of `hyper` :sweat_smile:

seanmonstar:
Yea, I allow this in hyper _knowing_ that `httparse` can never look at the values, but it would be better to be explicit about it. It would be a breaking change, that's true. I haven't spent much time thinking about how to fix it in this repo, and instead just promise myself that the slice can never be inspected :shrug:

Soveu:
Or we can just introduce a function that would just take buffer and parse immediately, it would fit for `hyper`
```rust
fn with_uninit_buffer_and_parse(headers: &mut [MaybeUninit<Header>]) -> Result<usize>
```
I have already adjusted `parse_headers_iter` to use this type

PR #97, opened 2021-05-11
Add methods for parsing with uninitialized buffers

Soveu:
I think it solves #95


Soveu:
I just realised this could be written more like the original code :/

Soveu:
Actually, I might have found a bug in my code :cold_sweat:

Issue #58, opened 2019-07-22
Private function shrink() is unsound


Shnatsel:
This code does not uphold Rust safety invariants: either `debug_assert!()` should be `assert!()` or the function must be marked `unsafe fn`:

https://github.com/seanmonstar/httparse/blob/6f696f5d027f35e11a70181c839b574e20335a74/src/lib.rs#L38-L43

Also, it's weird to see a custom function for this - `slice[..len]` looks like it should be sufficient, but I'm probably missing something.

seanmonstar:
Good call, the function should be `unsafe` for sure.

It's been so long, I can't remember why I wrote like this instead of with slice syntax. Do the lifetimes work out? It might have been to not need the bounds check, but I can't remember if that showed in the benchmarks.

cuviper:
You can't directly re-slice it, but you can disconnect the lifetimes with `mem::replace` like:

```rust
#[inline]
fn shrink<T>(slice: &mut &mut [T], len: usize) {
    let full = mem::replace(slice, &mut []);
    *slice = &mut full[..len];
}
```

Note that this will be bounds-checked, if that matters for your performance, and if it does panic the original will be left with the empty slice. I guess you could pre-assert the length to avoid that, and then hope that LLVM will elide the later check.

hkBst:
Merge #98 completely changed this code, including removing this function and its single use, so I think this issue can be closed. What do you think, @Shnatsel ?

PR #98, opened 2021-05-12
Uninit headers - second approach

Soveu:
Previous attempt: #97 
Fixes: #95 and #58


Soveu:
@seanmonstar do we consider bumping minimal rustc version, like in `hyper`? Looks like CI is failing (mainly) because of no `MaybeUninit`

Soveu:
@seanmonstar ?

RalfJung:
> Looks like CI is failing (mainly) because of no MaybeUninit in 1.10

https://crates.io/crates/maybe-uninit would let you support at least Rust 1.20 while ensuring soundness on Rust 1.36+.

Even older versions of Rust do not really have any good way to work with uninit data correctly. If they still need to be supported, possibly a feature flag could be used to provide this API only when the user asks for it? I will note though that Rust 1.20 is already ancient, it was released 4 years ago.

seanmonstar:
Ok, let's work on this! First thing I did was get CI moved from Travis to GH Actions.

For MSRV, it's probably fine we just bump it to 1.36. It's two years old at this point!

Soveu:
A force-pushed empty commit would restart CI?

seanmonstar:
Hm, probably a rebase of master.

Soveu:
ugh, sorry for the wait, git didn't want to cooperate with me -_-

seanmonstar:
Yea, so the new CI is failing the msrv job. You can just update the yaml file to set the new MSRV.

seanmonstar (approved):

PR #99, opened 2021-05-24
Print Header value as string if it's UTF-8

Thomasdezeeuw:
Rather then looking at a slice of bytes is easier (for most humans) to
read a string. Since most, but not all, header values are strings
attempt to create a string from the header value and print that, falling
back to printing it a slice of bytes if the value is not a valid UTF-8
string.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -504,7 +504,7 @@ impl<'h, 'b> Response<'h, 'b> {
 }
 
 /// Represents a parsed header.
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
Ping @seanmonstar.

nox:
@Thomasdezeeuw Could you force push to trigger the CI again?

Thomasdezeeuw:
@nox also rebase on master.

seanmonstar (approved):

PR #105, opened 2021-09-23
Fixes #70 - Update README to be consistent with docs example

### LINKED ISSUE #70: Please increase the demo number of headers
tldr: Please change the documentation demo line from
```
let mut headers = [httparse::EMPTY_HEADER; 16];
let mut req = httparse::Request::new(&mut headers);
```
to 
```
let mut headers = [httparse::EMPTY_HEADER; 64];
let mut req = httparse::Request::new(&mut headers);
```

---
This is a really silly bug that bit us.

Our server uses httparse. Since we weren't quite sure how many headers we needed to pre-allocate, we used the demo code's suggested 16 headers. Unfortunately, with work going on around the new `Sec-*` headers, we suddenly saw a spike of clients trying to connect with 17 headers. It took us a bit to figure out what was going on and why these connections were mysteriously failing. 

Boosting the count might help others not suddenly hit this problem.

miketaylr:
cf. https://docs.rs/httparse/1.3.4/httparse/constant.EMPTY_HEADER.html#example

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Works with `no_std`, simply disable the `std` Cargo feature.
 ## Usage
 
 ```rust
-let mut headers = [httparse::EMPTY_HEADER; 16];
+let mut headers = [httparse::EMPTY_HEADER; 64];
 let mut req = httparse::Request::new(&mut headers);
 
 let buf = b"GET /index.html HTTP/1.1\r\nHost";


miketaylr:
Please take a look @seanmonstar, thanks!

seanmonstar (approved):
Yup! Thanks

PR #106, opened 2021-12-30
Fix SIMD header value check on char >= 0x80

eaufavor:
The SIMD intrinsics *_cmpgt_epi8 are for signed chars.
This change correctly performs the unsigned comparison.


nox:
Could you describe the problem a bit more? Why does it matter that the intrinsics are for signed chars? Is there a bug in current httparse?

eaufavor:
The intention of the code is to check if the char is greater than 0x1f. However for a char >= 0x80, *_cmpgt_epi8 treats it as negative number. So all the chars that >= 0x80 are treated as if they are  <= 0x1f, which will make them to be rejected as illegal header values.

nox (approved):

seanmonstar (approved):

Issue #68, opened 2020-02-26
Should Line Folding get replaced by spaces?


spease:
Noticed this in RFC-7230:

>    A user agent that receives an obs-fold in a response message that is
>    not within a message/http container MUST replace each received
>    obs-fold with one or more SP octets prior to interpreting the field
>    value.

I think currently httparse will just reject a line fold. This paragraph seems to suggest it should reinterpret it with spaces. Is this interpretation correct?

I am running into this problem (and an invalid SOH character) with an internal server.

Thanks.

seanmonstar:
Yes, it "should" handle line folds. I've purposefully left it out for a few reasons:

- It's very very rare.
- It's deprecated, so new implementations should NEVER write it, and old ones should upgrade.
- Because of those, I didn't want the performance cost or code complexity of supporting it.

tari:
I've encountered this problem while working on a WARC parser, where that format [specifies definitions compatible with RFC2616](https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.0/#file-and-record-model). There's a [WARC proposal](https://github.com/iipc/warc-specifications/issues/74) to forbid implementers of a future version from writing folded lines, but readers would still need to be able to handle the old version for backwards compatibility.

---

I appreciate the complexity of supporting some of the weird corners of header syntax, but it's a shame if `httparse` never supports those since it means potential users will have to roll their own parsers if that functionality is important. If this effectively means that the library is designed for server use only that would be reasonable, but it would be good to document that limitation.

PR #107, opened 2022-01-13
Implement header line folding (fixes #37, #68)

### LINKED ISSUE #37: multiline headers not handled
While RFC 7230 deprecated mutliline headers (search around for `obs-fold` in the RFC, they're still something you sometimes encounter. I noticed this while using the multipart_mime library, which in turn uses httparse to handle headers in a MIME message. Here's a failing test case:

```
    req! {
        test_multiline_header,
        b"GET / HTTP/1.1\r\nX-Received: by 10.84.217.214 with SMTP id whatever;\r\n        Wed, 21 Jun 2017 09:04:21 -0700 (PDT)",
        |req| {
            assert_eq!(req.headers.len(), 1);
        }
    }
```

I'm not sure how to actually *fix* the issue, but I figured I'd at least report the bug.

nox:


src/lib.rs
```diff
@@ -1389,6 +1465,122 @@ mod tests {
...
+            .allow_obsolete_multiline_headers_in_responses(true)
+            .parse_response(&mut response, RESPONSE_WITH_OBSOLETE_LINE_FOLDING_IN_MIDDLE);
+
+        assert_eq!(result, Ok(Status::Complete(RESPONSE_WITH_OBSOLETE_LINE_FOLDING_IN_MIDDLE.len())));
+        assert_eq!(response.version.unwrap(), 1);
+        assert_eq!(response.code.unwrap(), 200);
+        assert_eq!(response.reason.unwrap(), "OK");
+        assert_eq!(response.headers.len(), 1);
+        assert_eq!(response.headers[0].name, "Line-Folded-Header");
+        assert_eq!(response.headers[0].value, &b"hello  \r\n \r\n there"[..]);
```
seanmonstar, line 1552: I see, this was what I was wondering: how exactly does it get exposed. Am I right in assuming that this implies it will "simply" capture the text from the first line until the end of all obs-folded lines? So, if this were `hello: there\r\n mister\r\n baguette\r\n`, it would scoop it all as `b"there\r\n mister\r\n baguette"`?

That seems elegant. Just to consider what I'd thought of, I assumed this feature would instead use up an entry in the `&mut [Header]` slice for each line. If the line started with obs-folding, then the parse would just grab the name from the previous entry. So it'd look something like this:

```rust
assert_eq!(resp.headers[0].name, "hello");
assert_eq!(resp.headers[0].value, &b"there");
assert_eq!(resp.headers[1].name, "hello");
assert_eq!(resp.headers[1].value, &b"mister");
assert_eq!(resp.headers[2].name, "hello");
assert_eq!(resp.headers[2].value, &b"baguette");
```

Whatever we end up using, explaining that is what it does would be useful as a sentence/paragraph in the commit description.

nox: Line folds are equivalent to spaces, but multiple entries of a single header are equivalent to a single entry combined with commas, so we can't return multiple entries.

seanmonstar: Oh right.... gross...

Welp, then I think this is the best we can do. I think it'd be good to better explain, perhaps with an example, in the documentation for that builder method.

seanmonstar (approved):
Looks great, thanks!

PR #112, opened 2022-03-18
Switch benches to criterion and remove pico targets

acfoltzer:
Criterion allows us to crank up the sampling size and confidence, which is important for these kinds of microbenchmarks that can have high variance. It also has other bells and whistles like tracking relative performance between runs, which can be very helpful for development, and no longer requires a nightly compiler to run the benchmarks.

Per discussion in Discord, the presence of pico in these benchmarks has outlived its usefulness; we're now primarily concerned with the relative performance of httparse between changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,14 @@ default = ["std"]
 std = []
 
 [dev-dependencies]
-pico-sys = "0.0.1"
+criterion = "0.3.5"
+
+[lib]
+bench = false
+
+[[bench]]
+name = "parse"
+harness = false
 
 [profile.bench]
 lto = true
--- a/benches/parse.rs
+++ b/benches/parse.rs
@@ -1,9 +1,9 @@
-#![feature(test)]
-
-extern crate pico_sys as pico;
+extern crate criterion;
 extern crate httparse;
 
-extern crate test;
+use std::time::Duration;
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion, Throughput};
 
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -18,7 +18,14 @@ default = ["std"]
 std = []
 
 [dev-dependencies]
-pico-sys = "0.0.1"
+criterion = "0.3.5"
+
+[lib]
+bench = false
```
acfoltzer, line 24: Not strictly necessary for this change, but gets a lot of extraneous output out of `cargo bench` invocations even when `--bench parse` is not specified.

benches/parse.rs
```diff
@@ -22,101 +22,72 @@ Keep-Alive: 115\r\n\
...
+    c.benchmark_group("resp_short")
+        .throughput(Throughput::Bytes(RESP_SHORT.len() as u64))
+        .bench_function("resp_short", |b| b.iter(|| {
+            assert_eq!(black_box(resp.parse(RESP_SHORT).unwrap()), httparse::Status::Complete(RESP_SHORT.len()));
+        }));
+}
+
+criterion_group!{
+    name = benches;
+    config = Criterion::default().sample_size(100).measurement_time(Duration::from_secs(10));
```
acfoltzer, line 90: This 10 second number can be tweaked if it slows down CI too much, but when spotting performance regressions for functions this small it's very helpful for keeping noise down.

seanmonstar (approved):

seanmonstar:
Awesome stuff, thank you! <3

PR #113, opened 2022-03-27
run Miri on CI

RalfJung:
This only works if we use the fallback implementation with Miri.

(FWIW, using `core::simd` would work, but `core::arch` simply has too many intrinsics to support them all in Miri.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
       - test
       - simd
       - msrv
+      - miri
     steps:
       - run: exit 0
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -127,3 +128,22 @@ jobs:
...
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          components: miri
+          override: true
+
+      - name: Test
+        run: MIRIFLAGS="-Zmiri-tag-raw-pointers -Zmiri-check-number-validity" cargo miri test
```
seanmonstar, line 149: I assume you'll know best: what's the likelihood of these changing and causing our CI to fail?

RalfJung: These flags basically put Miri into "hard mode", adding some extra checks that are off-by-default but give extra confidence in the unsafe code.  I'm not sure what kind of `unsafe` code httparse uses or which kind of code you plan to use in the future, so I cannot really give failure probabilities for the UB checks.

When we change flags we usually first just deprecate them (showing a warning) because we know removing them would break CI. Even this happens rarely. The last actual breaking change wrt our flags was https://github.com/rust-lang/miri/pull/1769; back then I made an effort to send PRs to all projects I could find that used Miri on CI to fix them preemptively. I don't even remember the last breaking flag change before that...

seanmonstar: Sounds good to me, thanks!

seanmonstar (approved):

Issue #83, opened 2021-01-25
double quotes in headers can not be parsed


guoyunlong16:
In the [header name map](https://github.com/seanmonstar/httparse/blob/master/src/lib.rs#L96), index 34 is false. So when parsing headers, the program will return Error::HeaderName when meeting double quotes(ascii number is 34) . However, bouble quotes in headers can be parsed correctly in chrome. Is this intentional or a bug?

Test Example:
```
 #[test]
    fn test_double_quotes() {
        use std::mem;
        let bytes= b"HTTP/1.1 200 OK\r\nServer: nginx/1.14.2\r\nDate: Mon, 25 Jan 2021 06:20:06 GMT\r\nContent-Type: image/png\r\nContent-Length: 24623\r\nConnection: keep-alive\r\n\"Access-Control-Allow-Origin: *\"\r\nAccept-Ranges: bytes\r\nAccess-Control-Allow-Origin: *\r\nCache-Control: 2592000\r\n\r\n";
        let mut headers: [Header; 10] = unsafe { mem::uninitialized() };
        let mut res = Response::new(&mut headers);
        let parsed_res = res.parse(bytes);
        println!("parsed res= {:?}", parsed_res);
    }
```

tari:
At least according to RFC 2616, Chrome's behavior seems correct. A header name is a sequence of `token`, and `token` is any character excluding `CTL`s or `SEPARATOR`s, neither of which includes `"`. However RFC 7230 (which obsoletes 2616) defines a `token` as a sequence of `tchar`, a class which specifically excludes double quotes.

I'd think a parser should be able to handle the old definition because it's impossible to tell whether a given message conforms to RFC 2616 or 7230 (both describe HTTP/1.1), but practically the new definition seems to have been changed because parsing the old one is much more complex than it initially seems and many implementations differ in their interpretation.

---

As a related example, RFC 7230 (section 3.2.4) deprecates line folding for header values and allows servers receiving folded lines to reject them, but also requires user agents to accept folded lines by converting to sequences of spaces before interpreting. In that particular instance, it seems like this library should implement the behavior specified for user agents because that is also permitted for servers.

tari:
It seems like the rationale in #68 applies here as well though: supporting the old behavior is difficult and probably slow, so it's simply not supported.

nox:
https://github.com/seanmonstar/httparse/pull/114 will let you ignore the invalid header, which is what Chrome is doing.

PR #114, opened 2022-04-22
Allow ignoring invalid header lines (fixes #61, #83)

### LINKED ISSUE #61: httparse should skip invalid headers
I am trying to use `reqwest` to parse a response from a server I don't control. `reqwest` uses `hyper` which uses `httparse` for parsing HTTP/1.x headers. Anyway, this server has a weird bug where it consistently returns a single corrupted header line in an otherwise completely valid response (the header contains unescaped non-token characters). Specifically, for some reason it tries to send the `DOCTYPE` as a header. The bug is unlikely to be fixed (this is old software), but it isn't really a problem because the page displays fine in all major browsers.

It seems that all major browsers simply ignore invalid header lines. However, `httparse` returns an error that aborts the entire parsing process. IMO this is a problem and should be fixed.

Here's a screenshot from Chrome that shows the invalid header being ignored:

ignored
original

In fact, Chrome's behavior is commented as: ["skip malformed header"](https://chromium.googlesource.com/chromium/src/+/b6532ab4077e913e38d99f5a970e437516643d03/net/http/http_util.cc#935).

Although technically changing this could be breaking, in this case, I can't imagine that any code would rely on response parsing to *fail* in this particular case.

Here are the relevant lines:

https://github.com/seanmonstar/httparse/blob/6f696f5d027f35e11a70181c839b574e20335a74/src/lib.rs#L594-L595

https://github.com/seanmonstar/httparse/blob/6f696f5d027f35e11a70181c839b574e20335a74/src/lib.rs#L613-L614

https://github.com/seanmonstar/httparse/blob/6f696f5d027f35e11a70181c839b574e20335a74/src/lib.rs#L672

https://github.com/seanmonstar/httparse/blob/6f696f5d027f35e11a70181c839b574e20335a74/src/lib.rs#L613-L614

I think all of these would be fixed by consuming `b` until the next newline then `continue 'headers`. I would open a PR but I just want to check that you agree that this change should be made.

nox:


seanmonstar:
It'd probably be better to have the separate commits as different PRs, as that would make the review panel a bit clear-er... But I looked at each commit individually, and I think it looks right.

@acfoltzer wanna take a look?

acfoltzer:
@seanmonstar sure thing, can you tag me as a reviewer so I don't lose track? Should be able to dig in early next week if that's alright

nox:
There is an actual bug now in httparse which makes invalid HTTP requests be accepted by the crate, could we hurry to get a new release?

Coder-256:
I wonder if the existing behavior is vulnerable to response splitting, e.g. `GET / HTTP/1.1\n\n: Foo\n\nBody`; if `config.allow_spaces_after_header_name_in_responses` is enabled, would `: Foo` be parsed as a header instead of part of the body?

nox:
Will do, but there isn't anything in this PR that should be long to review 🤷🏻‍♂️

nox:
https://github.com/seanmonstar/httparse/pull/115

nox:
> I wonder if the existing behavior is vulnerable to response splitting, e.g. `GET / HTTP/1.1\n\n: Foo\n\nBody`; if `config.allow_spaces_after_header_name_in_responses` is enabled, would `: Foo` be parsed as a header instead of part of the body?

No, because the check for the first char being a valid header name char is separate, but now that you say this, I just realised I forgot to handle the first char being invalid, in this PR. I guess there was some nasty stuff to review after all hah!

acfoltzer:
Thank you! 

> Will do, but there isn't anything in this PR that should be long to review 🤷🏻‍♂️

I want to make sure we have time to carefully understand the changes being made to the parser, and give some of my fuzzers time to chug on the new branch. After all, the impact if we get something wrong is pretty significant.

acfoltzer (changes requested):
Thank you again for spinning off #115, I've left my review over there.

As for the new proposed feature `ignore_invalid_header_lines_in_responses` in [7244d92](https://github.com/seanmonstar/httparse/pull/114/commits/7244d9266dfdead6b07e87d53eed479e0e7a6fe3), I have some concerns before we move ahead.

## Overall thoughts

Beside the two specific areas of concern described below, I'm also wondering whether this level of permissiveness belongs in this API. The interaction with `obs-fold` was one that happened to occur to me as I read the patch, but it would not surprise me if there are other potential interactions that could lead to undesirable behavior just within the parser, let alone downstream effects from clients or servers observing incomplete sets of headers without awareness of the ignored field(s). 

The other options we support for parsers at least have precedent in current or past versions of the HTTP/1 specs, but the proposed behavior steps entirely outside of those. Can you share more about the use case you have in mind for this? Are there clients and servers emitting such messages in practice today? Is this a hedge against data corruption? 

Even with a compelling use case, this flag should be very heavily caveated in documentation along with specific descriptions of the cases where its use is required.

## Scope of the new option

If we move ahead with adding this flag, the name should be changed and documentation added to more narrowly describe the behavior, or the implementation should expand to also allow ignoring errors that arise during parsing of the header value. In the current implementation, parse failures only lead to an ignored line if they occur at or before the colon separating the name from the value. Other errors that can arise after this point are not ignored. For example, modifying one of the new tests in 7244d92 to put the invalid character in the value does not yield the same behavior:

```rust
static RESPONSE_WITH_INVALID_CHAR_IN_HEADER_VALUE: &'static [u8] =
    b"HTTP/1.1 200 OK\r\nAccess-Control-Allow-Credentials: tr\x0Bue\r\nBread: baguette\r\n\r\n";

#[test]
fn test_ignore_header_line_with_invalid_char_in_value() {
    let mut headers = [EMPTY_HEADER; 2];
    let mut response = Response::new(&mut headers[..]);
    let result = ::ParserConfig::default()
        .allow_spaces_after_header_name_in_responses(true)
        .ignore_invalid_header_lines_in_responses(true)
        .parse_response(&mut response, RESPONSE_WITH_INVALID_CHAR_IN_HEADER_VALUE);

    // This assertion fails:
    assert_eq!(result, Ok(Status::Complete(81)));
    assert_eq!(response.version.unwrap(), 1);
    assert_eq!(response.code.unwrap(), 200);
    assert_eq!(response.reason.unwrap(), "OK");
    assert_eq!(response.headers.len(), 1);
    assert_eq!(response.headers[0].name, "Bread");
    assert_eq!(response.headers[0].value, &b"baguette"[..]);
}
```

## Interaction with `obs-fold`

Ignoring up to the following newline does not mix well with `allow_obsolete_multiline_headers_in_responses`:

```rust
static RESPONSE_WITH_OBSOLETE_LINE_FOLDING_AND_INVALID_NAME: &'static [u8] =
    b"HTTP/1.1 200 OK\r\nLine-Folded-Header\xFF: hello  \r\n \r\n there\r\nBread: baguette\r\n";

/// This test passes; obs-fold is accepted but the invalid header name is not
#[test]
fn test_forbid_response_with_obsolete_line_folding_in_middle_and_invalid_name() {
    let mut headers = [EMPTY_HEADER; 1];
    let mut response = Response::new(&mut headers[..]);
    let result = ::ParserConfig::default()
        .allow_obsolete_multiline_headers_in_responses(true)
        .parse_response(&mut response, RESPONSE_WITH_OBSOLETE_LINE_FOLDING_AND_INVALID_NAME);

    assert_eq!(result, Err(::Error::HeaderName));
}

/// This should pass if we're ignoring the invalid folded line, but does not
#[test]
fn test_ignore_folded_line_with_invalid_name() {
    let mut headers = [EMPTY_HEADER; 1];
    let mut response = Response::new(&mut headers[..]);
    let result = ::ParserConfig::default()
        .allow_obsolete_multiline_headers_in_responses(true)
        .ignore_invalid_header_lines_in_responses(true)
        .parse_response(&mut response, RESPONSE_WITH_OBSOLETE_LINE_FOLDING_AND_INVALID_NAME);

    assert_eq!(
        result,
        Ok(Status::Complete(
            RESPONSE_WITH_OBSOLETE_LINE_FOLDING_IN_MIDDLE.len()
        ))
    );
    assert_eq!(response.version.unwrap(), 1);
    assert_eq!(response.code.unwrap(), 200);
    assert_eq!(response.reason.unwrap(), "OK");
    assert_eq!(response.headers.len(), 1);
    assert_eq!(response.headers[0].name, "Bread");
    assert_eq!(response.headers[0].value, &b"baguette"[..]);
}
```

I don't believe this would lead to a situation where the second line of a folded line could be confused for a legit `field-line` all on its own, as a valid folded line must be preceded by `RWS` whitespace. But it's close enough that I would want to heavily scrutinize any logic that determines when the ignored header line is truly over.

nox:
> Beside the two specific areas of concern described below, I'm also wondering whether this level of permissiveness belongs in this API.

Yes it does, browsers do that so httparse should to.

<hr>

Will modify my patch to take care of errors in values.

<hr>

The interaction with obsolete line folding is a non-issue because additional lines must start with whitespace, and that's never considered a header name. I'll add a test which involves an invalid header that uses obsolete line folding though.

nox:
I also noticed yesterday that I didn't take care of header names whose very first char is invalid.

nox:
I've studied Chromium's code a bit more and the one thing it rejects is NUL chars. Everything else is ok.

src/lib.rs
```diff
@@ -983,44 +1047,25 @@ fn parse_headers_iter_uninit<'a, 'b>(
...
-                                    // empty.
-                                },
-                            }
-                        }
+                if b == b'\r' {
+                    b = next!(bytes);
+                }
+                if b != b'\n' {
+                    continue_with_new_header_on_next_line!(bytes, b, HeaderValue);
+                }
```
acfoltzer, line 1059: I think this changes the behavior so that a bare `\r` would be treated as a valid line ending while in the `whitespace_after_colon` loop, e.g., `"HTTP/1.1 200 OK\r\nFoo: \r\r\n"`

src/lib.rs
```diff
@@ -921,56 +942,99 @@ fn parse_headers_iter_uninit<'a, 'b>(
...
+                        // header or the final line return. This header is thus
+                        // empty.
+                    },
+                }
+            }
+        }
+    }
+
     'headers: loop {
+        macro_rules! continue_with_new_header_on_next_line {
```
acfoltzer, line 971: I like this macro but the name is confusing when viewed in calling contexts. It should emphasize that this yields the given error if `ignore_invalid_headers` is not enabled. To bikeshed, perhaps `error_or_ignore_header_line`?

src/lib.rs
```diff
@@ -361,6 +362,26 @@ impl ParserConfig {
...
+    /// You should only want this if you are implementing a proxy whose main
+    /// purpose is to sit in front of browsers whose users access arbitrary content
+    /// which may be malformed, and they expect everything that works without
+    /// the proxy to keep working with the proxy.
+    ///
+    /// This option will prevent `ParserConfig::parse_response` from returning
+    /// an error encountered when parsing a header, except if the error was caused
+    /// by the character NUL (ASCII code 0), as Chrome specifically always reject
+    /// those.
+    pub fn ignore_invalid_headers_in_responses(
```
acfoltzer, line 377: This option equally affects requests
```suggestion
    pub fn ignore_invalid_headers(
```

acfoltzer:
> Yes it does, browsers do that so httparse should to.
D'oh, of course browsers allow this :disappointed: 

> The interaction with obsolete line folding is a non-issue because additional lines must start with whitespace, and that's never considered a header name. I'll add a test which involves an invalid header that uses obsolete line folding though.

It looks like the new tests with folding address the problems I had in the original patch, nice!

> I've studied Chromium's code a bit more and the one thing it rejects is NUL chars. Everything else is ok.

I see that there's now more of a description of behavior in the doc comment, thank you! I think it'd be worth enumerating a few more details there, like how it determines the end of the line to ignore, and how it interacts with folding and possibly the other options. Not quite to the level of an RFC with ABNF etc, but it would be nice to be very specific about what this allows.

src/lib.rs
nox: It doesn't, because we pass `&ParserConfig::default()` to `parse_headers` in `parse_request`, that's what I was talking about on Discord yesterday.

acfoltzer: There are new methods though like `ParserConfig::parse_request()` that allow configuring the parser for requests. Existing behavior didn't change but new programs can use this behavior.

nox: Yes, and as I have stated multiple times, we pass `&ParserConfig::default()` to `parse_headers` in `parse_request`. The only `parse_request` function there is in that crate is the method on `ParseConfig`.

`ParserConfig::parse_request` calls `Request::parse_with_config` which calls `Request::parse_with_config_and_uninit_headers` which calls `parse_headers_iter_uninit` with `&ParserConfig::default()`, so all those flags related to header parsing don't apply to requests. That's a feature, because we don't want to relax request parsing, just response parsing. See https://github.com/seanmonstar/httparse/pull/91#discussion_r602604778.

src/lib.rs
nox: Fixed.

src/lib.rs
acfoltzer: Ah okay, I was missing the distinction between the config's new use in `parse_with_config_and_uninit_headers` here https://github.com/seanmonstar/httparse/blob/7fd85d187b180863013f75e38f3b4c344eafb9e4/src/lib.rs#L443-L449

And the fact that we call `parse_headers_iter_uninit` with the default config. That latter part is an oversight on my part when adding the multiple spaces flag for requests. Even though we don't currently have any flags that should override header parsing behavior in request _headers_, I think the inconsistency is surprising.

Is the argument against expanding support for `ignore_invalid_headers` to requests that your use case trusts the browser to generate valid requests? Other proxies might not be so lucky as to have that kind of constrained request input. 

I tend to agree that we shouldn't add support the new flag on requests until someone asks for it, but I think it's worth being more defensive than assuming we'll always pass `&ParserConfig::default()` to this function. It would be very easy for someone to make a subsequent change that requires customizing the config there without realizing that it would make `ignore_invalid_headers_in_responses` start applying equally to requests.

src/lib.rs
```diff
@@ -361,6 +362,39 @@ impl ParserConfig {
...
+    ///
+    /// This mimicks the behaviour of major browsers. You probably don't want this.
+    /// You should only want this if you are implementing a proxy whose main
+    /// purpose is to sit in front of browsers whose users access arbitrary content
+    /// which may be malformed, and they expect everything that works without
+    /// the proxy to keep working with the proxy.
+    ///
+    /// This option will prevent `ParserConfig::parse_response` from returning
+    /// an error encountered when parsing a header, except if the error was caused
+    /// by the character NUL (ASCII code 0), as Chrome specifically always reject
```
acfoltzer, line 375: Should also mention `\r` here?

src/lib.rs
```diff
@@ -361,6 +362,39 @@ impl ParserConfig {
...
+    /// those.
+    ///
+    /// The ignorable errors are:
+    /// * empty header names;
+    /// * characters that are not allowed in header names, except for `\0` and `\r`;
+    /// * when `allow_spaces_after_header_name_in_responses` is not enabled,
+    ///   spaces and tabs between the header name and the colon;
+    /// * missing colon between header name and colon;
```
acfoltzer, line 383:
```suggestion
    /// * missing colon between header name and value;
```

src/lib.rs
```diff
@@ -361,6 +362,39 @@ impl ParserConfig {
...
+    /// * characters that are not allowed in header values except for `\0` and `\r`.
+    ///
+    /// If an ignorable error is encountered, the parser tries to find the next
+    /// line in the input to resume parsing the rest of the headers. An error
```
acfoltzer, line 387: While the doc as a whole and implies that obs-fold header lines will be skipped once any part of them fails, it's probably worth calling that out explicitly.

acfoltzer (changes requested):
A couple small documentation tweaks, but then the remaining issue is whether to be more defensive about making this only apply to responses as described in https://github.com/seanmonstar/httparse/pull/114#discussion_r865171776

src/lib.rs
nox: Not exactly here, but I can mention it in a separate sentence yeah.

I reject `\r` because both Firefox and Chrome accept it, except in a different way (Firefox sees a plain old header value char, Chrome sees a header separator).

src/lib.rs
nox: I'll add a test that enables a response-only flag and checks that it doesn't apply to requests.

src/lib.rs
acfoltzer: Fun! I guess we can wait to find out whether someone will have a need for the FF or Chrome-specific behaviors. Maybe those could be extra flags in the future? Rejecting seems like the right call for now.

nox:
@acfoltzer Improved the docs and added a test for request parsing.

acfoltzer (approved):
Looking good to me now, thank you! As a followup, I will try to think about how best to make requests fully configurable without breaking the current response-only flags.

seanmonstar (approved):

PR #126, opened 2022-11-19
Allow passing custom configuration to `parse_headers`

gootorov:
There are several protocols that look almost identical to HTTP (in my case, it is SIP). Often the only difference is the `request-line` and/or the `response-line`. In SIP in particular, multiline headers are the norm ([allow_obsolete_multiline_headers_in_responses](https://docs.rs/httparse/latest/httparse/struct.ParserConfig.html#method.allow_obsolete_multiline_headers_in_responses)).

Being able to parse the `request/response-line` by myself and then using `parse_headers` with a custom configuration would be very convenient.


gootorov:
Friendly ping! :)
If you're unable to review for the time being, that's totally fine of course. Only making sure this isn't forgotten.

gootorov:
@seanmonstar Apologies for pinging again. Do you mind having a look at this PR? This is a small, backward-compatible change. Having this merged would be really great, if possible.

If there's something that needs to be changed, such as documentation, tests, etc -- please let me know.

PR #132, opened 2023-04-14
cleanup: SIMD runtime detection

AaronO:
Also cleanup, builds off #131

We can see the overhead improvements in uri parsing for smaller values (where overhead is relatively significant) and we can see it compound in header/count accumulating the overhead of jumping in & out of SIMD.

## header/count

|         | 1   | 2   | 4   | 8   | 16  | 32  | 64  | 128 |
|---------|-----|-----|-----|-----|-----|-----|-----|-----|
| Before | 22  | 39  | 77  | 144 | 283 | 578 | 1092| 2159|
| After  | 21  | 37  | 71  | 135 | 271 | 568 | 1025| 2034|

## uri

|         | 1b  | 2b  | 4b  | 8b  | 16b | 32b | 64b | 128b | 256b | 512b | 1024b | 2048b | 4096b |
|---------|-----|-----|-----|-----|-----|-----|-----|------|------|------|-------|-------|-------|
| Before | 7   | 8   | 9   | 11  | 8   | 6   | 7   | 11   | 19   | 34   | 67    | 127   | 270   |
| After  | 5   | 5   | 7   | 9   | 6   | 5   | 6   | 9    | 20   | 31   | 60    | 119   | 255   |

### DIFF
--- a/src/simd/avx2.rs
+++ b/src/simd/avx2.rs
@@ -1,29 +1,24 @@
 use crate::iter::Bytes;
 
-pub enum Scan {
-    /// Returned when an implementation finds a noteworthy token.
-    Found,
-    /// Returned when an implementation couldn't keep running because the input was too short.
-    TooShort,
-}
-
 #[cfg(target_arch = "x86")]
-unsafe fn parse_uri_batch_32(_: &[u8]) -> usize {
+pub unsafe fn match_uri_vectored(_: &mut Bytes) {
     unreachable!("AVX2 detection should be disabled for x86");
 }
 
+#[inline]
 #[cfg(target_arch = "x86_64")]
-#[target_feature(enable = "avx2")]
-pub unsafe fn parse_uri_batch_32(bytes: &mut Bytes) -> Scan {
+#[target_feature(enable = "avx2", enable = "sse4.2")]
+pub unsafe fn match_uri_vectored(bytes: &mut Bytes) {
     while bytes.as_ref().len() >= 32 {
         let advance = match_url_char_32_avx(bytes.as_ref());
         bytes.advance(advance);
 
         if advance != 32 {
-            return Scan::Found;
+            return;
         }
     }
-    Scan::TooShort
+    // do both, since avx2 only works when bytes.len() >= 32
+    super::sse42::match_uri_vectored(bytes)
 }
 
 #[inline(always)]
@@ -69,22 +64,23 @@ unsafe fn match_url_char_32_avx(buf: &[u8]) -> usize {
 }
 
 #[cfg(target_arch = "x86")]
-unsafe fn match_header_value_batch_32(_: &[u8]) -> usize {
+pub unsafe fn match_header_value_vectored(_: &mut Bytes) {
     unreachable!("AVX2 detection should be disabled for x86");
 }
 
 #[cfg(target_arch = "x86_64")]
-#[target_feature(enable = "avx2")]
-pub unsafe fn match_header_value_batch_32(bytes: &mut Bytes) -> Scan {
+#[target_feature(enable = "avx2", enable = "sse4.2")]
+pub unsafe fn match_header_value_vectored(bytes: &mut Bytes) {
     while bytes.as_ref().len() >= 32 {
         let advance = match_header_value_char_32_avx(bytes.as_ref());
         bytes.advance(advance);
 
         if advance != 32 {
-            return Scan::Found;
+            return;
         }
     }
-    Scan::TooShort
+    // do both, since avx2 only works when bytes.len() >= 32
+    super::sse42::match_header_value_vectored(bytes)
 }
 
 #[inline(always)]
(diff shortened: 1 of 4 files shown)


seanmonstar:
cc @Noah-Kennedy

AaronO:
Small enum in lieu of func ptr is marginally better thanks to branch-prediction, observed on header/count:

```
test header/count_1 ... bench:          21 ns/iter (+/- 5)
test header/count_2 ... bench:          35 ns/iter (+/- 5)
test header/count_4 ... bench:          66 ns/iter (+/- 2)
test header/count_8 ... bench:         130 ns/iter (+/- 53)
test header/count_16 ... bench:         259 ns/iter (+/- 80)
test header/count_32 ... bench:         499 ns/iter (+/- 43)
test header/count_64 ... bench:         978 ns/iter (+/- 195)
test header/count_128 ... bench:        1938 ns/iter (+/- 116)
```

src/simd/runtime.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+static mut RUNTIME_FEATURE: Option<RuntimeFeature> = None;
+
+pub fn match_uri_vectored(bytes: &mut Bytes) {
+    // SAFETY: calls are guarded by a feature check
+    unsafe {
+        if let None = RUNTIME_FEATURE {
+            RUNTIME_FEATURE = Some(RuntimeFeature::detect());
+        }
```
seanmonstar, line 30: This kind of unsynchronized access of a static could cause a data race if the code is called at the same time in separate threads.

AaronO: To my understanding this is correct, no thread/core should will perform a torn read or write given that we're storing/loading naturally aligned values of less than a quadword. [Ref AMD64 manual (7.3.2)](https://www.amd.com/system/files/TechDocs/24593.pdf#page=260):

> Cacheable, naturally-aligned single loads or stores of up to a quadword are atomic on any processor
model, as are misaligned loads or stores of less than a quadword that are contained entirely within a
naturally-aligned quadword.

Obviously there are other issues with shared globals, but in this instance it's a simple init once, worst case multiple threads will write, but it's idempotent so there's no observable issue.

The previous implementation which stored a shared pointer should also have been correct as long as the global is aligned. In this instance both `RuntimeFeature` and `Option<RuntimeFeature>` are a single byte which is always contained entirely with a naturally-aligned quadword.

These assumptions hold true:
```rs
#[test]
fn test_assumptions() {
    assert_eq!(std::mem::size_of::<RuntimeFeature>(), 1);
    assert_eq!(std::mem::size_of::<Option<RuntimeFeature>>(), 1);
    assert_eq!(RuntimeFeature::Avx2 as u8, 0);
    assert_eq!(RuntimeFeature::Sse42 as u8, 1);
    assert_eq!(RuntimeFeature::Nop as u8, 2);
    assert_eq!(unsafe { std::mem::transmute::<Option::<RuntimeFeature>, u8>(None) }, 3);
}
```

But I agree it's probably more idiomatic and obviously correct to wrap this in an atomic with a relaxed store which I believe will be codegen'd to single `mov`s anyway for the above reasons

AaronO: Made the change

seanmonstar: You're right. Sorry, I should have said more in my initial comment. On most architectures, it would be a non-tearing read/write. I had just checked in the Rust Nomicon, to see about the language semantics. I wanted to know, can we rely on that, or is there a theoretical instance it would be wrong. It makes no such promise, just claims these this is "unsound":

> - two or more threads concurrently accessing a location of memory
> - one or more of them is a write
> - one or more of them is unsynchronized

src/simd/avx2.rs
```diff
@@ -1,28 +1,27 @@
 use crate::iter::Bytes;
 
-pub enum Scan {
-    /// Returned when an implementation finds a noteworthy token.
-    Found,
-    /// Returned when an implementation couldn't keep running because the input was too short.
-    TooShort,
+#[cfg(target_arch = "x86")]
```
seanmonstar, line 3: What requires defining these stubs? It looks like the method signatures are different... But that would be a compile error, so are these ever even built?

AaronO: The sig mismatch is a mistake, I forgot to update it as I shifted the interface to be `math_x_vectored`. We indeed don't build it in our tests, with neon support it could be helpful to arm64/arm CI possibly x86 whilst we're at it.

The stub is "required" because it's expected by the interface and both x86/x64 paths expect avx2 functions to exist but then be guarded by the feature check. I don't see any reason why we could not just skip avx2 on x86

I'll address that in a follow up if you're onboard

AaronO:
@seanmonstar Squashed to a single commit [`cleanup: simd runtime detection`](https://github.com/seanmonstar/httparse/pull/132/commits/e7f4a8499df24561eff792595b975dc0483353a9), since it's more of a cleanup than a perf improvement as we reverted to the atomic (which shouldn't be an issue in absolute but I would rather fine tune minimizing overhead of runtime feature detection in a separate PR)

seanmonstar:
I know when I originally added SIMD support to this crate, the `is_x86_feature_detected!` macro did not get inlined, so the function call was slower than caching in an atomic locally. Inline attributes were later added, so it could be that the cache is no longer worth keeping. Would be good to measure.

seanmonstar (approved):

PR #135, opened 2023-04-22
mark match_uri_vectored/match_header_value_vectored function as unsafe

kitcatier:
https://github.com/seanmonstar/httparse/blob/fbb0bdde72a9b571ded779fdc3544d8a39c812fc/src/simd/mod.rs#L78-L87
hello, if a function's entire body is unsafe, the function is itself unsafe and should be marked appropriately. Marking them unsafe also means that callers must make sure they know what they're doing.

### DIFF
--- a/src/simd/mod.rs
+++ b/src/simd/mod.rs
@@ -75,13 +75,13 @@ pub use self::runtime::*;
 ))]
 mod sse42_compile_time {
     #[inline(always)]
-    pub fn match_uri_vectored(b: &mut crate::iter::Bytes<'_>) {
+    pub unsafe fn match_uri_vectored(b: &mut crate::iter::Bytes<'_>) {
         // SAFETY: calls are guarded by a compile time feature check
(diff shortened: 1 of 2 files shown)


seanmonstar:
I believe the point there is that they are unsafe if you call them on the wrong CPU architecture. But the `#[cfg]` attributes on the module above make it so those calls only happen if it is compiled for the _right_ CPU. So, that is why they can be considered safe. Sound right @AaronO?

AaronO:
> I believe the point there is that they are unsafe if you call them on the wrong CPU architecture. But the `#[cfg]` attributes on the module above make it so those calls only happen if it is compiled for the _right_ CPU. So, that is why they can be considered safe. Sound right @AaronO?

Yes this is 100% safe as per the `SAFETY:` comments. Functions marked with `#[target_feature(...)]` must be `unsafe` on x86/x64/aarch64 because you're codegening specialized instructions, but it's perfectly safe to call those functions once you've asserted (at compiletime or runtime) that those instructions are supported on the host CPU.

PR #137, opened 2023-04-24
simd: drop 2 unnecessary instructions

AaronO:
## Benches

```
before:
test header/value_256b ... bench:          34 ns/iter (+/- 13)
test header/value_512b ... bench:          50 ns/iter (+/- 5)
test header/value_1024b ... bench:          86 ns/iter (+/- 7)
test header/value_2048b ... bench:         172 ns/iter (+/- 20)
test header/value_4096b ... bench:         298 ns/iter (+/- 62)

after:
test header/value_256b ... bench:          33 ns/iter (+/- 10)
test header/value_512b ... bench:          48 ns/iter (+/- 20)
test header/value_1024b ... bench:          77 ns/iter (+/- 8)
test header/value_2048b ... bench:         149 ns/iter (+/- 8)
test header/value_4096b ... bench:         252 ns/iter (+/- 50)
```

### DIFF
--- a/src/simd/avx2.rs
+++ b/src/simd/avx2.rs
@@ -108,10 +108,9 @@ unsafe fn match_header_value_char_32_avx(buf: &[u8]) -> usize {
     let tab = _mm256_cmpeq_epi8(dat, TAB);
     let del = _mm256_cmpeq_epi8(dat, DEL);
     let bit = _mm256_andnot_si256(del, _mm256_or_si256(low, tab));
-    let rev = _mm256_cmpeq_epi8(bit, _mm256_setzero_si256());
-    let res = _mm256_movemask_epi8(rev) as u32;
-
-    res.trailing_zeros() as usize
+    let res = _mm256_movemask_epi8(bit) as u32;
+    // TODO: use .trailing_ones() once MSRV >= 1.46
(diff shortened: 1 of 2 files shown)


seanmonstar:
MSRV, currently 1.32, is grumpy. Looks like this needs Rust 1.46, which came out in August 2020, almost 3 years ago.

Distros are on newer versions already. 👍

AaronO:
> MSRV, currently 1.32, is grumpy. Looks like this needs Rust 1.46, which came out in August 2020, almost 3 years ago.
> 
> Distros are on newer versions already. 👍

I think an MSRV bump would be reasonable, especially since `hyper`@`master` is on `1.56`, but since we've managed without it so far we could mint this first series of perf changes in a release then bump MSRV and land this + other cleanups it enables. Bumping to 1.46 would allow for some slightly more idiomatic code in `swar` and other areas.

seanmonstar:
I would make the next release a minor version anyways, 1.9, since the addition of neon at least feels significant enough to not be a patch version. I'd be fine with an MSRV bump in 1.9, or we could do it in 1.10...

ankit-gautam23 (approved):
LGTM

seanmonstar (approved):

PR #138, opened 2023-04-24
refactor: Bytes inner arithmetic

AaronO:
IMO cleaner pointer arithmetic avoiding unnecessary intermediate slices.

No perf gains in isolation but facilitates some (upcoming PRs)

This is conceptually easier to reason about, end is fixed, "start" and "cursor" advance monotonically

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -1,3 +1,6 @@
+use core::convert::TryInto;
+use core::convert::TryFrom;
+
 #[allow(missing_docs)]
 pub struct Bytes<'a> {
     start: *const u8,
@@ -48,17 +51,11 @@ impl<'a> Bytes<'a> {
     }
     
     #[inline]
-    pub fn peek_n<U>(&self, n: usize) -> Option<U> {
+    pub fn peek_n<'b: 'a, U: TryFrom<&'a [u8]>>(&'b self, n: usize) -> Option<U> {
+        // TODO: once we bump MSRC, use const generics to allow only [u8; N] reads
         // TODO: drop `n` arg in favour of const
         // let n = core::mem::size_of::<U>();
-        // Boundary check then read array from ptr
-        if self.len() >= n {
-            let ptr = self.cursor as *const U;
-            let x = unsafe { core::ptr::read_unaligned(ptr) };
-            Some(x)
-        } else {
-            None
-        }
+        self.as_ref().get(..n)?.try_into().ok()
     }
 
     #[inline]
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -1,97 +1,144 @@
-use core::slice;
-use core::convert::TryInto;
-use core::convert::TryFrom;
-
 #[allow(missing_docs)]
 pub struct Bytes<'a> {
-    slice: &'a [u8],
-    pos: usize
+    start: *const u8,
+    end: *const u8,
+    cursor: *const u8,
+    phantom: core::marker::PhantomData<&'a ()>,
 }
 
 #[allow(missing_docs)]
 impl<'a> Bytes<'a> {
     #[inline]
     pub fn new(slice: &'a [u8]) -> Bytes<'a> {
+        let start = slice.as_ptr();
+        let end = unsafe { start.add(slice.len()) };
+        let cursor = start;
         Bytes {
-            slice,
-            pos: 0
+            start,
+            end,
+            cursor,
+            phantom: core::marker::PhantomData,
         }
     }
 
     #[inline]
     pub fn pos(&self) -> usize {
-        self.pos
+        self.cursor as usize - self.start as usize
     }
 
     #[inline]
     pub fn peek(&self) -> Option<u8> {
-        self.peek_ahead(0)
+        if self.cursor < self.end {
+            // SAFETY: bounds checked
+            Some(unsafe { *self.cursor })
+        } else {
+            None
+        }
     }
 
     #[inline]
     pub fn peek_ahead(&self, n: usize) -> Option<u8> {
-        self.slice.get(self.pos + n).copied()
+        let ptr = unsafe { self.cursor.add(n) };
+        if ptr < self.end {
+            // SAFETY: bounds checked
+            Some(unsafe { *ptr })
+        } else {
+            None
+        }
     }
-
+    
     #[inline]
-    pub fn peek_n<U: TryFrom<&'a[u8]>>(&self, n: usize) -> Option<U> {
-        self.slice.get(self.pos..self.pos + n)?.try_into().ok()
+    pub fn peek_n<U>(&self, n: usize) -> Option<U> {
+        // TODO: drop `n` arg in favour of const
+        // let n = core::mem::size_of::<U>();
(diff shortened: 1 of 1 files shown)


src/iter.rs
```diff
@@ -1,108 +1,158 @@
...
+    
     #[inline]
-    pub fn peek_n<U: TryFrom<&'a[u8]>>(&self, n: usize) -> Option<U> {
-        self.slice.get(self.pos..self.pos + n)?.try_into().ok()
+    pub fn peek_n<U>(&self, n: usize) -> Option<U> {
+        // TODO: drop `n` arg in favour of const
+        // let n = core::mem::size_of::<U>();
+        // Boundary check then read array from ptr
+        if self.len() >= n {
+            let ptr = self.cursor as *const U;
```
seanmonstar, line 56: I think this change makes the method no longer expose the right safety. The `n` and `U` are no longer related, and copying out an arbitrary `U` _could_ allow the wrong thing to happen, like if you were to peek something that implements `Drop` (and that thing does something with the memory). Before, since it was `TryFrom<&[u8]>`, the `U` had to be a safe copy.

src/iter.rs
```diff
@@ -1,108 +1,158 @@
...
+            end,
+            cursor,
+            phantom: core::marker::PhantomData,
         }
     }
 
     #[inline]
     pub fn pos(&self) -> usize {
-        self.pos
+        self.cursor as usize - self.start as usize
```
seanmonstar, line 26: This and the other arithmetic below in `len` don't end up costing more than just reading the integer? Or perhaps the other operations cost less, and they balance out?

src/iter.rs
AaronO: In #139 I update the code to drop `n` arg and use `core::mem::size_of::<U>()`, I didn't want to change the signature so this PR was a simple drop-in API-compatible reimplementation of `Bytes`

AaronO: `peek_n()` is really just used to read `[u8; N]`s, on more recent versions of `rust` we would express this with const generics. Both this signature and the previous versions are much wider than intended due to MSRV tradeoffs.

seanmonstar: Looking in #139, while it does tie it to the `n`, it's still possible to materialize an item that would be bad. For instance, if there were enough bytes left, `let boom = bytes.peek_n::<String>();` would be possible and safe. Which would be unsound.

The existing impl could be kept, creating a slice from the pointers and then using `TryFrom`. We could keep it safe with const-generics later.

AaronO: `TryFrom` ultimately ends up doing extra bounds-checking (hence why it's fallible), it's roughly something like:
```rs
fn try_from(slice: &[T]) -> Result<&[T; N], TryFromSliceError> {
    if slice.len() == N {
        let ptr = slice.as_ptr() as *const [T; N];
        // SAFETY: ok because we just checked that the length fits
        unsafe { Ok(&*ptr) }
    } else {
        Err(TryFromSliceError(()))
    }
}
```

These changes were collectively intended to allow for ~optimal bounds checking in hotpaths.

> For instance, if there were enough bytes left, `let boom = bytes.peek_n::<String>();` would be possible and safe. Which would be unsound.
> 
> The existing impl could be kept, creating a slice from the pointers and then using `TryFrom`. We could keep it safe with const-generics later.

I'm not 100% sure what we're trying to guard against here, `Bytes` and `peek_n` aren't publicly exposed to end-users. Technically someone could `peek_n` an array of bytes and transmute that to a struct (`String` ...)

I think your core point is that `peek_n` in its current form is `unsafe` and should be tagged as such ? (which I agree with in absolute)

My conclusion here is that we should likely just bump MSRV to `1.56` (or `1.51` for const generics), `1.51` is over 2y  old, we're unnecessarily over-constraining ourselves.

seanmonstar: > aren't publicly exposed

Yea, true, I know I'm being somewhat pedantic. But I prefer that functions, even internal ones, that are marked as safe are indeed sound in all ways. That way we don't make mistakes in the future when we forget why (or a new contributor who doesn't know the impl).

Agreed that removing the bounds checks would be good. I'm find with a bump to 1.51. (Is there anything decent that we'd need that is newer? I try to only upgrade MSRV if we're actually going to use it.)

seanmonstar: But, also, if you wanted to wait on the MSRV bump, the `TryFrom` wouldn't be any _worse_ than what is there already.

AaronO: Yeah I'll revert this to the `TryFrom` impl and leave a TODO for const generics

src/iter.rs
AaronO: This PR does enable perf improvements (just not without other changes in `lib.rs`  (and `simd/*`) to match the more optimal usage patterns this unblocks). On `req/req` (M1 Air) this takes us from `~114ns/iter` to `~70ns/iter` which is rather significant.

`.pos()` is infrequently read, it's not in the hotpath and #139 mostly stops using it because it's somewhat quirky.

Similarly `.len()` isn't required or optimal for bounds checking, for example in SIMD batches `.len()` mutates with each loop as you shorten the buffer. So whilst a slice `.len()` is just a read, it's crucial to notice that the actual slice (and thus `.len()`) we're checking is constantly changing since `bytes.as_ref()` is computing a new slice

What you really want is just not to overrun the buffer and perform an OOB read, so what you can do is compute `max(start, end - SIMD_BATCH_SIZE)` and bounds check that the current `ptr` is less than that. That `endsimd_ptr` is constant across an entire simd batch and broadly an entire parse call so the compiler can compute it once at top of loop keep it in a register whereas `.len` will constantly mutate (with every single `.advance()` call)

The core thing to understand with this PR is the following: `(*const u8, *const u8)` and `(*const u8, usize)` can both equally denote a slice or contiguous chunk of memory. In our case we're modelling a `Cursor` over that `slice`, let's look at those structs and their invariants as they mutate over time:

```rs
// This PR
struct PtrCursor {
  start: *const u8, // Increases monotonically
  cursor: *const u8, // Increases monotonically 
  end: *const, // Constant, never changes post initialization
}

// master
struct SliceCursor {
  slice_start: *const u8, // Increases monotonicallyx
  slice_len: usize, // Decreases monotonically
  pos: usize, // Increases, then is reset 
}
```

Conceptually `PtrCursor` is much easier to reason about (and optimize), especially when you note that it's basically:
```
struct PtrCursorFull {
    buf_start: *const u8,
    buf_end: *const u8,
}
```

src/iter.rs
```diff
@@ -1,108 +1,155 @@
...
+            Some(unsafe { *self.cursor })
+        } else {
+            None
+        }
     }
 
     #[inline]
     pub fn peek_ahead(&self, n: usize) -> Option<u8> {
-        self.slice.get(self.pos + n).copied()
+        let ptr = unsafe { self.cursor.add(n) };
```
hkBst, line 44: This unsafe lacks a safety comment, and is in fact unsound, as there is no guarantee that the safety conditions for `add` as explained [here](https://doc.rust-lang.org/std/primitive.pointer.html#method.add) are satisfied.

PR #134, opened 2023-04-20
refactor: simd swar

AaronO:
This refactor moves the block-wise validators to a "swar" SIMD backend

The core logic of validate => extract => chain is (IMO) now more evident
(and how we stack validators, using the largest SIMD validator available then finishing with scalar if at end of buffer [uncommon])

Perf wise, this is roughly on par with master, slightly faster on some regards but not 100% fine-tuned or fully benched with neon or avx/sse. In part because it avoids duplicating validating work between SIMD and scalar validators.

PR #139, opened 2023-04-24
refactor: cleaner header parsing

AaronO:
Builds off #134 (swar), #138 (Bytes cursor)

Cleaner, faster and less macros !

## Key changes

- Broke down header-parsing into clean conceptual steps (whilst being faster !)
- Added InnerResult allowing for idiomatic `?` early-exits, removing need for parsing-helper macros
- Removed macros.rs, leaving only `byte_map!` (response header-parsing macros should become functions)

### TODO

- [ ] convert response header-parser, supporting its quirks


seanmonstar:
Happy to see those gnarly macros cleaned up. I put those together when I was very new at Rust, and slowly grew them poorly, and understanding them whenever I came back to them was always confusing.

I'm guessing moving around the enums makes them slightly smaller. You said there was a measurable perf improvement?

seanmonstar:
Just checking in here, did you want to get this merged before we shoot out the new release?

PR #133, opened 2023-04-15
perf: simd neon

AaronO:
First pass at neon support, building off #132

I shared some promising benches here: https://github.com/seanmonstar/httparse/pull/123#issuecomment-1509870871

Overall it's a net-positive but not totally optimal.

Next steps we would want to land #123 and then refactor it into a SWAR SIMD backend and stack them appropriately to avoid duplicated work and cleaner logic.

Additionally, offsetz is very much central to total perf, I believe the vectorized implementation I added underperforms due to inefficiently reloading the offset lookup table instead of keeping it a register at the top of the header parsing loop.

PR #141, opened 2023-04-25
fix: neon no_std

AaronO:
I accidentally typed `std::mem` out of habit, this wasn't caught in #133 because `neon+no_std` isn't exercised

### DIFF
--- a/src/simd/neon.rs
+++ b/src/simd/neon.rs
@@ -165,8 +165,8 @@ unsafe fn offsetz(x: uint8x16_t) -> u32 {
 unsafe fn offsetnz(x: uint8x16_t) -> u32 {
     // Extract two u64
     let x = vreinterpretq_u64_u8(x);
-    let low: u64 = std::mem::transmute(vget_low_u64(x));
-    let high: u64 = std::mem::transmute(vget_high_u64(x));
+    let low: u64 = core::mem::transmute(vget_low_u64(x));
+    let high: u64 = core::mem::transmute(vget_high_u64(x));
 
     #[inline]
     fn clz(x: u64) -> u32 {
--- a/src/simd/neon.rs
+++ b/src/simd/neon.rs
@@ -165,8 +165,9 @@ unsafe fn offsetz(x: uint8x16_t) -> u32 {
 unsafe fn offsetnz(x: uint8x16_t) -> u32 {
     // Extract two u64
     let x = vreinterpretq_u64_u8(x);
-    let low: u64 = core::mem::transmute(vget_low_u64(x));
-    let high: u64 = core::mem::transmute(vget_high_u64(x));
+    // Extract to general purpose registers to perform clz
+    let low: u64 = vgetq_lane_u64::<0>(x);
+    let high: u64 = vgetq_lane_u64::<1>(x);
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
I was digging through the docs (I wanted to check on the transmute), and noticed it wasn't [stabilized until 1.59](https://doc.rust-lang.org/src/core/stdarch/crates/core_arch/src/arm_shared/neon/mod.rs.html#5259)? Woops, oh well :)

seanmonstar:
So, it looks it's an `uint64x1_t`, which doesn't have a `repr(transparent)` on it. Is there it publicly documented somewhere that we can transmute it into the primitive? That's not usually allowed (transmuting a unit struct to the field isn't guaranteed to be right).

AaronO:
> So, it looks it's an `uint64x1_t`, which doesn't have a `repr(transparent)` on it. Is there it publicly documented somewhere that we can transmute it into the primitive? That's not usually allowed (transmuting a unit struct to the field isn't guaranteed to be right).

[The docs](https://doc.rust-lang.org/core/arch/aarch64/struct.uint64x1_t.html) describe it as:
> ARM-specific 64-bit wide vector of one packed u64.

[The code](https://doc.rust-lang.org/src/core/stdarch/crates/core_arch/src/arm_shared/neon/mod.rs.html#19-106) is:
```rs
/// ARM-specific 64-bit wide vector of one packed `u64`.
#[cfg_attr(not(target_arch = "arm"), stable(feature = "neon_intrinsics", since = "1.59.0"))]
pub struct uint64x1_t(pub(crate) u64);
```

It's obviously correct. They `transmute` in their own tests to `u64 => uint64x1_t`, but I guess the idiomatic way would be [vget_lane_u64](https://doc.rust-lang.org/core/arch/aarch64/fn.vget_lane_u64.html) should compile to the same code moving the values from the SIMD registers to general purpose registers.

AaronO:
> I was digging through the docs (I wanted to check on the transmute), and noticed it wasn't [stabilized until 1.59](https://doc.rust-lang.org/src/core/stdarch/crates/core_arch/src/arm_shared/neon/mod.rs.html#5259)? Woops, oh well :)

Yeah, I guess we could compile-time check for `neon_intrinsics` or the rust version guard on that and fallback to SWAR ?
(Rust 1.59 is 1yo, still not bad but understand why you want to be conservative)

seanmonstar:
Yea, I know it works transmuting it now, but the [transmute docs](https://doc.rust-lang.org/nomicon/transmutes.html) point out it's not "defined", so it can't be depended on:

> For repr(C) types and repr(transparent) types, layout is precisely defined. But for your run-of-the-mill repr(Rust), it is not.

`repr(simd)` isn't really documented there at all. This also includes structs with a single field, they technically could have a different layout than the field alone, even if they currently don't. 🤷 

> (Rust 1.59 is 1yo, still not bad but understand why you want to be conservative)

The biggest "old" one would be Debian, which is really close to that range. Some other environments have even older compilers.

AaronO:
@seanmonstar See #142, added a test for aarch64 on MSRv and guarded it so we fallback to `swar` if < 1.59

PR #144, opened 2023-06-01
Enabled the use of clippy for the project.

f0rki:
* Run clippy in CI
    - Require SAFETY documents in front of every unsafe block.
    - Set MSRV version in clippy.toml to avoid lints that break the MSRV
* Several fixes to make clippy happy
    - Ran cargo clippy --fix
    - Added some clippy exceptions
    - Added many SAFETY comments to unsafe blocks
* Tested MIRAI static analyzer to find "unintentional panics"
    - added some debug-build overflow checks to satsify MIRAI
    - Currently, installing MIRAI takes way too long in CI so it is
      disabled for now. Need to find a way to cache this (maybe a custom github action or docker container).


.github/workflows/ci.yml
```diff
@@ -158,6 +159,16 @@ jobs:
...
 
+  clippy_check:
+    runs-on: ubuntu-latest
+    # Make sure CI fails on all warnings, including Clippy lints
+    env:
+      RUSTFLAGS: "-Dwarnings"
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run Clippy
+        run: cargo clippy --all-targets --all-features
```
seanmonstar, line 170: Can we make this just check a few conservative lints? I don't want to block builds for _all_ clippy warnings.

f0rki: I removed the `-Dwarnings` env variable, which doesn't fail CI on warnings. The default clippy `deny`s seem reasonable. I added the `deny` undocumented unsafety to `src/lib.rs`

seanmonstar (approved):

PR #145, opened 2023-08-04
Ignoring invalid headers for requests

grnmeira:
The idea for this PR came up from [this](https://github.com/hyperium/hyper/discussions/3267) discussion in the hyper repo.

This PR tries to make a request version of the `ParserConfig::ignore_invalid_headers_in_responses` configuration, introduced as `ParserConfig::ignore_invalid_headers_in_requests`, so that we can also ignore invalid headers in requests.

In order to achieve that avoiding breaking backwards compatibility, it splits the `ParserConfig` struct used for `parse_headers_iter_uninit` into `ParserConfig` and `HeaderParserConfig`.

Happy to reassess the tests or the approach.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,4 +1,4 @@
-///! Utility macros
+//! Utility macros
 
 macro_rules! next {
     ($bytes:ident) => ({
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -413,6 +413,7 @@ impl ParserConfig {
         self
     }
 
+    /// Sets whether invalid header lines should be silently ignored in requests.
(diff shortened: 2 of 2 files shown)


src/macros.rs
```diff
@@ -1,4 +1,4 @@
-///! Utility macros
+//! Utility macros
```
grnmeira, line 1: Not sure why clippy started acting up with this now. The lint was introduced on 1.70 and latest commit is the one enabling clippy for the pipelines.

grnmeira:
@seanmonstar any chances we could get this in? Happy to make any changes if needed too.

seanmonstar (approved):

PR #155, opened 2024-02-08
add option to save partial header value

traceflight:
Hello @seanmonstar,

Thanks for this great crate. I'm using this as the basic parser for http data analysis. I want to get as much as information even though the http date is not complete. So i create this PR to add an option to save partial header value. 

Hope you can review this code.


traceflight:
#155

PR #156, opened 2024-02-28
simd: split cursor advancing from value matching

lucab:
This refactors all SIMD modules in order to make the value-matching logic self-contained. Thus, all bytes-cursor manipulations are now grouped and performed once at the end, outside of SIMD logic.

Performance impact on my AVX2-capable workstation seems positive (arbitrary benchmark-noise filtering at >20%):

critcmp

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -953,18 +953,20 @@ fn parse_token<'a>(bytes: &mut Bytes<'a>) -> Result<&'a str> {
 #[allow(missing_docs)]
 // WARNING: Exported for internal benchmarks, not fit for public consumption
 pub fn parse_uri<'a>(bytes: &mut Bytes<'a>) -> Result<&'a str> {
-    let start = bytes.pos();
-    simd::match_uri_vectored(bytes);
     // URI must have at least one char
-    if bytes.pos() == start {
+    let uri_len = simd::match_uri_vectored(bytes.as_ref());
+    if uri_len == 0 {
         return Err(Error::Token);
     }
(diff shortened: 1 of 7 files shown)


lucab:
@seanmonstar this is ready for a review pass, whenever you have time.

There is a minor cleanup bundled in this PR (marking several functions as `pub(crate)`) which I did in order to make sure I wasn't changing public APIs. I can split that to a dedicated PR if you prefer.

I'll be honest, I started doing this rework as part of https://github.com/hyperium/hyper/issues/3574 before actually going for https://github.com/hyperium/hyper/pull/3575, focused on memory usage/allocation patterns.
My goal was to (eventually) use the `bytes` crate in this library, but then I realized that this required SIMD-related groundwork was actually providing performance improvements on its own. As such, I think it makes sense to land this already.

seanmonstar (approved):
Beautiful PR, and the speed boosts seem out of this world!

lucab:
Thanks for merging this. Even if I recorded those perf numbers myself, I'm still somehow puzzled and a bit skeptical about them.
I tried on a different machine (an AMD Ryzen 7) and this time I'm not seeing any measurable change (i.e. all differences are below 20%).
Both cases are workstation laptops and I did disable all kind of turbo-boosting and CPU dynamic scaling I could think of.
So either the code change is reliably hitting something very specific on the Intel machine (better cache locality?), or this workstation is just a-not-very good environment for benchmark comparisons.

Overall, I think the new code is a useful refactor but I personally won't guarantee the pictured performance changes to be the valid in all environments.

PR #157, opened 2024-02-28
lib: fix import

lucab:
This fixes the style of an import statement.


lucab:
Note: this is a cosmetic PR which can be quickly landed, so that I can re-trigger CI on https://github.com/seanmonstar/httparse/pull/156 without having to wait for approval on each iteration.

seanmonstar (approved):

PR #160, opened 2024-04-11
Resize headers on partial parse

tgnottingham:
The main benefit of this is that it enables you to get the parsed headers from partial results when using uninitialized headers. Prior to this, on a partial result, the headers in a request or response would be replaced with the original slice passed when creating the request or response (typically an empty slice when using uninitialized headers), rather than the initialized portion of the uninitialized headers passed into the parse function call.

It also makes handling partial parse results more convenient in general since you can iterate over the headers in the request or response without having to manually determine the last element.

This is technically a breaking change, though it seems unlikely that any client would depend on the previous behavior in a way that the change would break their code.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -547,7 +547,7 @@ impl<'h, 'b> Request<'h, 'b> {
         newline!(bytes);
 
         let len = orig_len - bytes.len();
-        let headers_len = complete!(parse_headers_iter_uninit(
+        let parse_headers_status = parse_headers_iter_uninit(
             &mut headers,
             &mut bytes,
             &HeaderParserConfig {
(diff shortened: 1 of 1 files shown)


tgnottingham:
I realized that there is a use case where a client might have depended on the previous behavior -- repeatedly using the same Request/Response until the result is Complete. With this change, if a later parse input has more header fields than a previous input, it will result in a TooManyHeaders error because the previous partial parse shrunk the headers.

Of course, even without my change, Complete Requests/Responses can't be reused for the exact same reason. Still, I'm guessing the change in behavior in the Partial case is enough to dissuade accepting this PR. Bummer.

PR #163, opened 2024-05-04
Add optional max header list size

hjr3:
Prior to parsing headers, both the request and response parsers now check that the remaining buffer length does not exceed the configured max size. If the length does exceed the max size, a new HeadersTooLarge error is returned.

Relates to https://github.com/hyperium/hyper/issues/3283 and https://github.com/hyperium/hyper/pull/3637#discussion_r1584576533


seanmonstar:
Hm, at this level, I would probably assume the caller could enforce this limit. Since this library just parses headers from a slice that the caller owns.

hjr3:
> Hm, at this level, I would probably assume the caller could enforce this limit. Since this library just parses headers from a slice that the caller owns.

My thinking is that we want to avoid spending time parsing the headers if we already know the the header list size exceeds the max. In order to do so, we need to know the remaining length of the buffer after that initially parsing the request/response.

If we are fine accepting the parsing time, then we could enforce the limit post-parsing at https://github.com/hyperium/hyper/blob/c62ea805b5b7ecdd56d69106387048218b4521ba/src/proto/h1/role.rs#L180

PR #174, opened 2024-06-11
Fix no space after colon

### LINKED ISSUE #172: Weird header parsing in 1.9
Thank you so much for `httparse` all your other amazing libraries.

Sorry for the maybe bad bug report, I am not a HTTP expert.

With the new 1.9 version I get a test regression in [oxhttp](https://github.com/oxigraph/oxhttp) with strange parsing result. A `:` gets included in one of the header value.

Here is a test to reproduce the possible regression:

```rust
    #[test]
    fn test_http_parse() {
        let mut headers = [httparse::EMPTY_HEADER; 1024];
        let mut parsed_request = httparse::Request::new(&mut headers);
        parsed_request
            .parse(b"POST / HTTP/1.1\ncontent-length:4\n\n")
            .unwrap();
        for header in parsed_request.headers {
            if header.name == "content-length" {
                assert_eq!(header.value, b"4"); // Fails, we get ':4'
            }
        }
    }
```

Thank you again for your library.

### LINKED ISSUE #173: httparse 1.9.x breaks header parsing for certain cases
A bug in the latest release leads to header values being incorrectly parsed when there is no space between header-name and header-value (`header-name:headervalue\r\n` includes `:` in the header value).

example:
```
    let mut headers = [httparse::EMPTY_HEADER; 1];
    let mut response = httparse::Response::new(&mut headers[..]);
    response.parse(b"HTTP/1.0 200 OK\r\nfoo:bar\r\n\r\n").unwrap();
    println!("{:?}", headers);
```

```
[Header { name: "foo", value: ":bar" }]
```

a quick bisect shows that https://github.com/seanmonstar/httparse/commit/b2625f38cc3a200e840861f2a3371eb8f5afdd73 introduced the regression.

seanmonstar:
It was noticed with a bisect the bug was introduced in this commit, so I'm reverting it for now to get a fix out faster.

cc @lucab 

Fixes #172 
Fixes #173

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -953,20 +953,18 @@ fn parse_token<'a>(bytes: &mut Bytes<'a>) -> Result<&'a str> {
 #[allow(missing_docs)]
 // WARNING: Exported for internal benchmarks, not fit for public consumption
 pub fn parse_uri<'a>(bytes: &mut Bytes<'a>) -> Result<&'a str> {
+    let start = bytes.pos();
+    simd::match_uri_vectored(bytes);
     // URI must have at least one char
-    let uri_len = simd::match_uri_vectored(bytes.as_ref());
-    if uri_len == 0 {
+    if bytes.pos() == start {
         return Err(Error::Token);
     }
-    // SAFETY: these bytes have just been matched here above.
-    unsafe { bytes.advance(uri_len) };
-    let uri_slice = bytes.slice();
 
-    let space_delim = next!(bytes);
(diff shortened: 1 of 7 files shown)


lucab:
@seanmonstar sorry for the trouble, I did miss a cursor-advancing instruction in one spot. Thanks for landing a testcase for this. I've opened https://github.com/seanmonstar/httparse/pull/175 which contains a plain reapply of your revert, plus a fixing commit.

Issue #146, opened 2023-08-14
support raw UTF8 paths


lovasoa:
Hello !
  I discovered recently that many user agents send raw UTF8 bytes in HTTP paths, and proxies forward them without issues. See https://github.com/actix/actix-web/issues/3102

  However, it looks like this library fails to parse such HTTP requests, preventing web servers from handling them.

  Would it be possible to handle such requests in this library ?

seanmonstar:
I feel like the bytes that httparse initially allowed were from an RFC, but I no longer remember which one it was (7230? URIs? WHATWG Fetch?)

lovasoa:
But if actual user agents, including safari, do send requests with utf8 paths, shouldn't we support them in server-side libraries, even if there is an RFC saying they are invalid ?

seanmonstar:
The general answer to that question is not always easy. There's sometimes good reasons to prevent them.

In this case, it could be fine to make the parser more relaxed. I was trying to remember what said to make it strict originally, so I could read if they included any comments as to why.

JakeChampion:
I believe this is coming from [rfc3986 section 3.3](https://www.rfc-editor.org/rfc/rfc3986.html#section-3.3) and the definition for a path segment not allowing non-us-ascii characters:

```abnf

   path          = path-abempty    ; begins with "/" or is empty
                 / path-absolute   ; begins with "/" but not "//"
                 / path-noscheme   ; begins with a non-colon segment
                 / path-rootless   ; begins with a segment
                 / path-empty      ; zero characters

   path-abempty  = *( "/" segment )
   path-absolute = "/" [ segment-nz *( "/" segment ) ]
   path-noscheme = segment-nz-nc *( "/" segment )
   path-rootless = segment-nz *( "/" segment )
   path-empty    = 0<pchar>

   segment       = *pchar
   segment-nz    = 1*pchar
   segment-nz-nc = 1*( unreserved / pct-encoded / sub-delims / "@" )
                 ; non-zero-length segment without any colon ":"

   pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"


   pct-encoded   = "%" HEXDIG HEXDIG

   unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```

[...]

PR #178, opened 2024-08-29
feat(path): allow utf8 chars in path

joelwurtz:
Ref https://github.com/seanmonstar/httparse/issues/146

I'm not used to work with simd parser so i have done what i think is right but not sure if it's right in all use case.

Mainly the uri non compliant rfc3986 parser is the same as the current one but allow every char between 128 to 255 which sould make it utf8 compliant, if i recall correctly code point always are above or equal to 128

I did not make a feature but rather an option in parser, however i wonder if this should be true by default ? or if it should be a feature (enabled by default ?) ?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2185,6 +2185,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(miri, ignore)] // Miri is too slow for this test
     fn test_all_utf8_char_in_paths() {
         // two code points
         for i in 128..256 {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,35 +63,26 @@ fn is_method_token(b: u8) -> bool {
     }
 }
 
-// ASCII codes to accept URI string.
-// i.e. A-Z a-z 0-9 !#$%&'*+-._();:@=,/?[]~^
+// char codes to accept URI string.
+// i.e. b'!' <= char and char != 127
 // TODO: Make a stricter checking for URI string?
 static URI_MAP: [bool; 256] = byte_map![
     0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-//  \0                            \n
     0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-//  commands
     0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-//  \w !  "  #  $  %  &  '  (  )  *  +  ,  -  .  /
-    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 0, 1,
-//  0  1  2  3  4  5  6  7  8  9  :  ;  <  =  >  ?
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-//  @  A  B  C  D  E  F  G  H  I  J  K  L  M  N  O
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-//  P  Q  R  S  T  U  V  W  X  Y  Z  [  \  ]  ^  _
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-//  `  a  b  c  d  e  f  g  h  i  j  k  l  m  n  o
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0,
-//  p  q  r  s  t  u  v  w  x  y  z  {  |  }  ~  del
-//   ====== Extended ASCII (aka. obs-text) ======
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
 ];
 
 #[inline]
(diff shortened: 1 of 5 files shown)


joelwurtz:
After playing with this, i think it should be a feature instead of a config, and it should be enabled by default.

Even if it's not really compliant with the RFC there seems to be some browsers that include utf8 in path of a HTTP request and without this it failed, having a feature allow someone to create a strict server but IMO default use case should handle most of web browser.

I also added a new Error if the given utf8 in path is not valid

src/lib.rs
```diff
@@ -963,6 +989,14 @@ pub fn parse_uri<'a>(bytes: &mut Bytes<'a>) -> Result<&'a str> {
             return Err(Error::Token);
         }
 
+        #[cfg(feature = "utf8_in_path")]
+        // SAFETY: all bytes up till `i` must have been `is_token` and therefore also utf-8.
+        return match str::from_utf8(unsafe { bytes.slice_skip(1) }) {
```
joelwurtz, line 994: This induce an overhead of 30 to 40% perf in uri benches, so not sure if we want this ? but in this case we are not sure that the string is valid utf8 unless using it (see the relative test)

src/lib.rs
```diff
@@ -963,6 +989,14 @@ pub fn parse_uri<'a>(bytes: &mut Bytes<'a>) -> Result<&'a str> {
...
+            Ok(uri) => Ok(Status::Complete(uri)),
+            Err(_) => Err(Error::Utf8Error),
```
joelwurtz, line 996: I created a specific error but it could also be a TokenError if we don't want to add too much burden for existing lib

src/lib.rs
```diff
@@ -2053,7 +2087,7 @@ mod tests {
         assert_eq!(parse_chunk_size(b"567f8a\rfoo"), Err(crate::InvalidChunkSize));
         assert_eq!(parse_chunk_size(b"567f8a\rfoo"), Err(crate::InvalidChunkSize));
         assert_eq!(parse_chunk_size(b"567xf8a\r\n"), Err(crate::InvalidChunkSize));
-        assert_eq!(parse_chunk_size(b"ffffffffffffffff\r\n"), Ok(Status::Complete((18, std::u64::MAX))));
+        assert_eq!(parse_chunk_size(b"ffffffffffffffff\r\n"), Ok(Status::Complete((18, u64::MAX))));
```
joelwurtz, line 2090: unrelated, can remove it, but without this test without default features would fail (since we don't have std)

src/lib.rs
joelwurtz:
```
uri/uri_0001b            1.00      1.1±0.01ns  1786.2 MB/sec    4.58      4.9±0.09ns   390.1 MB/sec
uri/uri_0002b            1.00      1.4±0.01ns     2.1 GB/sec    4.45      6.0±0.34ns   474.7 MB/sec
uri/uri_0004b            1.00      1.9±0.02ns     2.5 GB/sec    4.36      8.1±0.08ns   590.5 MB/sec
uri/uri_0008b            1.00      1.2±0.08ns     6.8 GB/sec    4.98      6.2±0.10ns  1390.9 MB/sec
uri/uri_0016b            1.00      1.8±0.01ns     9.0 GB/sec    2.80      4.9±0.08ns     3.2 GB/sec
uri/uri_0032b            1.00      2.8±0.20ns    11.1 GB/sec    2.05      5.7±0.12ns     5.4 GB/sec
uri/uri_0064b            1.00      5.5±0.75ns    11.1 GB/sec    1.30      7.1±0.04ns     8.5 GB/sec
uri/uri_0256b            1.11     19.1±0.09ns    12.5 GB/sec    1.00     17.2±2.03ns    14.0 GB/sec
uri/uri_0512b            1.36     40.4±0.26ns    11.8 GB/sec    1.00     29.8±3.51ns    16.0 GB/sec
uri/uri_1024b            1.35     80.6±0.53ns    11.8 GB/sec    1.00     59.9±0.34ns    15.9 GB/sec
uri/uri_2048b            1.69    163.7±1.62ns    11.7 GB/sec    1.00     97.1±0.45ns    19.7 GB/sec
uri/uri_4096b            1.95   310.9±22.99ns    12.3 GB/sec    1.00    159.4±0.92ns    23.9 GB/sec
version/http11           1.05      0.8±0.05ns    11.9 GB/sec    1.00      0.7±0.00ns    12.5 GB/sec
```

joelwurtz: I also try to use [simdutf8](https://github.com/rusticstuff/simdutf8/), but not sure we wants a dep for that, result are better on large string, but no really significant on small string : 

```
group                    no_utf8                                simd_utf8                              utf8
-----                    -------                                ---------                              ----
uri/uri_0001b            1.00      1.1±0.01ns  1786.2 MB/sec    5.42      5.8±0.07ns   329.6 MB/sec    4.58      4.9±0.09ns   390.1 MB/sec
uri/uri_0002b            1.00      1.4±0.01ns     2.1 GB/sec    4.80      6.5±0.02ns   440.1 MB/sec    4.45      6.0±0.34ns   474.7 MB/sec
uri/uri_0004b            1.00      1.9±0.02ns     2.5 GB/sec    4.34      8.0±0.44ns   593.1 MB/sec    4.36      8.1±0.08ns   590.5 MB/sec
uri/uri_0008b            1.00      1.2±0.08ns     6.8 GB/sec    5.58      6.9±0.06ns  1240.3 MB/sec    4.98      6.2±0.10ns  1390.9 MB/sec
uri/uri_0016b            1.00      1.8±0.01ns     9.0 GB/sec    3.47      6.1±0.07ns     2.6 GB/sec    2.80      4.9±0.08ns     3.2 GB/sec
uri/uri_0032b            1.00      2.8±0.20ns    11.1 GB/sec    2.48      6.9±0.12ns     4.5 GB/sec    2.05      5.7±0.12ns     5.4 GB/sec
uri/uri_0064b            1.00      5.5±0.75ns    11.1 GB/sec    1.23      6.7±0.13ns     9.0 GB/sec    1.30      7.1±0.04ns     8.5 GB/sec
uri/uri_0128b            1.21     10.3±0.06ns    11.7 GB/sec    1.00      8.5±1.10ns    14.1 GB/sec    1.19     10.2±0.90ns    11.8 GB/sec
uri/uri_0256b            1.72     19.1±0.09ns    12.5 GB/sec    1.00     11.1±0.30ns    21.5 GB/sec    1.54     17.2±2.03ns    14.0 GB/sec
uri/uri_0512b            2.02     40.4±0.26ns    11.8 GB/sec    1.00     20.0±0.36ns    23.9 GB/sec    1.49     29.8±3.51ns    16.0 GB/sec
uri/uri_1024b            2.48     80.6±0.53ns    11.8 GB/sec    1.00     32.5±1.46ns    29.4 GB/sec    1.84     59.9±0.34ns    15.9 GB/sec
uri/uri_2048b            2.70    163.7±1.62ns    11.7 GB/sec    1.00     60.6±1.06ns    31.5 GB/sec    1.60     97.1±0.45ns    19.7 GB/sec
uri/uri_4096b            2.77   310.9±22.99ns    12.3 GB/sec    1.00    112.4±1.09ns    33.9 GB/sec    1.42    159.4±0.92ns    23.9 GB/sec

**Comment by seanmonstar** [comment]
I asked in the referenced issue to check with a few implementers if this is desirable :)

File: Cargo.toml [hunk 4]
```diff
@@ -13,7 +13,8 @@ edition = "2018"
 build = "build.rs"
 
 [features]
-default = ["std"]
+default = ["std", "utf8_in_path"]
+utf8_in_path = []
```
seanmonstar, line 17: Let's not bother with a Cargo feature, it'll just be allowed, since I can tell from operators and also looking in other libraries, this is a thing that just happens.

src/lib.rs
```diff
@@ -161,6 +182,9 @@ pub enum Error {
     TooManyHeaders,
     /// Invalid byte in HTTP version.
     Version,
+    #[cfg(feature = "utf8_in_path")]
+    /// Invalid UTF-8 in path.
+    Utf8Error,
```
seanmonstar, line 187: Unfortunately we cannot add a new error variant without a breaking change. But it's probably correct to just consider this a `Token` error.

src/lib.rs
```diff
@@ -2161,7 +2195,7 @@ mod tests {
         assert_eq!(result, Err(crate::Error::Token));
     }
 
-    static REQUEST_WITH_MULTIPLE_SPACES_AND_BAD_PATH: &[u8] = b"GET   /foo>ohno HTTP/1.1\r\n\r\n";
+    static REQUEST_WITH_MULTIPLE_SPACES_AND_BAD_PATH: &[u8] = b"GET   /foo ohno HTTP/1.1\r\n\r\n";
```
seanmonstar, line 2198: We should make sure to keep a unit test for DEL, perhaps it should be this one? With a code comment above reminding us that a DEL test should be kept.

seanmonstar (commented):
Thanks! I think we can do this, but some changes commented inline.

Also, to test all SIMD variants, we can probably just have a unit test that loops all possible bytes of URI_MAP, making a long enough vec with that byte, and then checking the match value is the same as without SIMD. I believe we have a test like that for one of the other byte maps.

joelwurtz:
I have apply your comments and squash, will add a test later that will try to iterate over all utf8 char so we are sure to not miss something

joelwurtz:
I added the est for this, is this what you want or did i misunderstood ?

joelwurtz:
Hey, do you need something more for this to be reviewed or merged ?

earce-pulse:
+1 looking forward to progress on this!

lovasoa:
@seanmonstar , do you think this can be merged ! This would really help us over at [SQLPage](https://github.com/sqlpage/SQLpage).

earce-pulse:
@seanmonstar us as well

joelwurtz:
I rebased the PR, i see that the other PR in hyperium was merged, let me know if this PR still need more work

seanmonstar (approved):

lovasoa:
thank you!

PR #182, opened 2024-09-02
simd: reduce pub functions visibility

lucab:
This tweaks all SIMD `pub` functions, moving them to `pub(crate)` instead.

### DIFF
--- a/src/simd/avx2.rs
+++ b/src/simd/avx2.rs
@@ -2,7 +2,7 @@ use crate::iter::Bytes;
 
 #[inline]
 #[target_feature(enable = "avx2", enable = "sse4.2")]
-pub unsafe fn match_uri_vectored(bytes: &mut Bytes) {
+pub(crate) unsafe fn match_uri_vectored(bytes: &mut Bytes) {
     while bytes.as_ref().len() >= 32 {
         let advance = match_url_char_32_avx(bytes.as_ref());
         bytes.advance(advance);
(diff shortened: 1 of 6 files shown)


AaronO:
What's the broader rationale ?

Wouldn't this "violate" [redundant_pub_crate](https://rust-lang.github.io/rust-clippy/master/index.html#/redundant_pub_crate) ?

seanmonstar:
FWIW, I am not convinced by the description at the clippy lint at all.

> Why is this bad?
>
> Writing pub(crate) is misleading when it’s redundant due to the parent module’s visibility.

There's nothing misleading about it. It doesn't cause any harm.

AaronO:
> There's nothing misleading about it. It doesn't cause any harm.

You could say it doesn't "harm", but in this instance it doesn't solve or address anything either ...
(I'd consider it if it served some clear ulterior motive, but at face value seems like a pointless change)

I'd argue it's somewhat a responsibility leak... Do we care about inner interface visibility much or predominantly the one exposed by `lib.rs` ? `mod simd;` already covers this, it's redundant IMO.

lucab:
Apologies, I should've added some breadcrumbs upfront; I've update the PR description with some refs now.

I think this `rustc` builtin lint [description](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.UNREACHABLE_PUB.html) describes it better than what I could do with my own words. Unfortunately it is not enforced by default due to historical reasons (`pub(_)` did not exist until 1.18).
I personally did this locally to lower the cognitive load, not having to check for possible API breakages in the middle of other refactors.

I can't speak for the clippy lint, however I did notice that it is in `nursery` (it has been [demoted](https://github.com/rust-lang/rust-clippy/pull/5372) there 4 years ago) and with a bunch of [open issues](https://github.com/rust-lang/rust-clippy/issues?q=is%3Aissue+redundant_pub_crate+is%3Aopen) and people voicing for its [removal](https://github.com/rust-lang/rust-clippy/issues/5369#issuecomment-687702757).

PR #185, opened 2024-09-25
fix unsound use of pointer::add

krasimirgg:
No functional changes intended.

From [the docs of pointer::add](https://doc.rust-lang.org/std/primitive.pointer.html#method.add): 
>If the computed offset is non-zero, then self must be derived from a pointer to some [allocated object](https://doc.rust-lang.org/std/ptr/index.html#allocated-object), and the entire memory range between self and the result must be in bounds of that allocated object. In particular, this range must not “wrap around” the edge of the address space.


seanmonstar:
Thanks! Notably, this was reported in #177.

Thoughts @AaronO?

seanmonstar (approved):

PR #187, opened 2024-10-22
iter::slice_skip remove possibility of underflow in debug_assert

hkBst:
`self.cursor.sub(skip)` must point inside the allocation of this object or this is UB, not to mention the possibility that if `skip` is very large self.cursor.sub(skip) may underflow and fail to trigger the debug_assertion.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -109,7 +109,7 @@ impl<'a> Bytes<'a> {
     /// implies a skip of at most 3).
     #[inline]
     pub unsafe fn slice_skip(&mut self, skip: usize) -> &'a [u8] {
-        debug_assert!(self.cursor.sub(skip) >= self.start);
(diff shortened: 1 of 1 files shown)


hkBst:
Apparently `offset_from` stabilized in 1.47.0 (35 versions behind 1.82.0 times 6 weeks is 210 weeks or about 4 years) is too new as MSRV=1.36. It "is equivalent to (self as isize - origin as isize) / (mem::size_of::<T>() as isize), except that it has a lot more opportunities for UB, in exchange for the compiler better understanding what you are doing.", but I suppose I can make it work...

hkBst:
MSRV is now 1.47.0, and this patch now passes all checks. There should not be any performance impact, since this only changes a debug_assert.

seanmonstar (approved):

PR #188, opened 2024-10-22
Inline peek_ahead to fix 186 and remove iter::peek_ahead as it is now unused.

hkBst:


hkBst:
Actually, is there really any reason for this complicated dance? Why not just match on "POST "? It seems that would probably do fewer bounds checks as well. Or if i does one for GET and then another one for POST, maybe it would work to match on `bytes.len()`?

seanmonstar:
I can say that method was very carefully adjusted to improve performance significantly.

hkBst:
Right, well this patch just inlines what the compiler was asked to inline, possibly a slightly more efficient version, so there should be no performance regression.

Is there anything you dislike about this approach?

seanmonstar:
Hm, could we just make `peek_ahead` unsafe and revert back to using `cursor.add(n)` once, adding a comment that the caller must be sure `n` is within the range? I think that'd keep the one place it's called a little cleaner.

PR #190, opened 2024-10-26
Fix method parsing

jeddenlea:
The `is_token` function, used exclusively for parsing the method in a request line, allows more values than it should. In particular, it allows a leading space to be parsed.  This problem is not exposed in hyper, which revalidates any method extracted by httparse, otherwise I'm sure this would have been noticed sooner!

Checking for a single range of valid bytes is very fast, so I've taken care to make sure that making `is_token` more complicated doesn't slow down the most common case.  While exploring a variety of options, I found the existing benchmark scheme to be a bit misleading because it would test only a single method at a time, so I've made a new benchmark that roughly simulates a mix of requests.  Ultimately, what I found to be a reasonable fix without any slowdown for the 99.9999% case is to check `b'A'..=b'Z'` and then fall back to a "byte map".

Both methods and header names have the same set of allowed bytes, a "token", but their uses are slightly different. I thought it would make sense to rename `is_token` to `is_method_token`, to mimic `is_header_name_token`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,7 +44,7 @@ pub mod _benchable {
     pub use super::iter::Bytes;
 }
 
-/// Determines if byte is a token char.
+/// Determines if byte is a method token char.
 ///
 /// > ```notrust
 /// > token          = 1*tchar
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -55,8 +55,11 @@ pub mod _benchable {
 /// >                ; any VCHAR, except delimiters
 /// > ```
 #[inline]
-fn is_token(b: u8) -> bool {
-    b > 0x1F && b < 0x7F
+fn is_method_token(b: u8) -> bool {
+    match b {
+        b'A'..=b'Z' => true,
```
seanmonstar, line 60: So that I'm sure I understood what you were saying, this range match is inlined here because it's faster _if the match is true_ than reaching into the token map? (That'd make sense, I assume this is compiles to a couple cmps, vs having to load from the map that's perhaps not in cache.)

If so, would you mind adding a little comment here that it's inline for perf reason? So Future Us doesn't go "what's this, debleeted". :)

jeddenlea: Certainly!

seanmonstar (approved):
Thanks!

PR #191, opened 2024-11-19
Fix unsoundness of peek_ahead in iter.rs

hkBst:
This fixes the inadequate safety checks, by checking the index `n` against the length of the byte array. Then there is no need to use wrapping_add instead of add since we now never point outside the array (since we checked the index against length already).

Potentially, there is some performance impact here, but the benchmarks are not very clear about the direction. Perhaps there is none to speak of?

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -1,5 +1,5 @@
-use core::convert::TryInto;
 use core::convert::TryFrom;
+use core::convert::TryInto;
 
 #[allow(missing_docs)]
 pub struct Bytes<'a> {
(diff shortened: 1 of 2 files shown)


seanmonstar:
Checking in Godbolt, it seems like this produces less optimized assembly code. Could we just mark the function unsafe and pass the requirement onto the caller?

hkBst:
I'd love to do that, but I don't see how it would help:

The single [caller lib::parse_method](https://github.com/seanmonstar/httparse/blob/380f130d0e5c7269e887b250a3ddcfcbfabfaaec/src/lib.rs#L847) does:
 - a safe get for 4 bytes `match bytes.peek_n::<[u8; 4]>(4) {`
 - then wants to look at the 5th byte `Some(POST) if bytes.peek_ahead(4) == Some(b' ') => {`
 - then unsafely advances 5 bytes `bytes.advance(5);`

So, we need to know that the 5th byte exists, which means we need to check against the length of the Bytes somehow.

hkBst:
Right, I suppose that allows us to simplify the in-bounds check to just a comparison with bytes.end(), but it would also lead to a baroque safety requirement and make `peek_ahead` an extremely single-use abstraction. I'll try and whip something up.

hkBst:
@seanmonstar what do you think of this version?

seanmonstar (approved):
Perfect! Thanks for keeping up on this <3

PR #205, opened 2025-02-26
fix(swar): utf8 support

### LINKED ISSUE #201: SWAR checks for carets while other SIMD implementations don't
Hi, first of all, thank you for this library. Learned a lot while reading through it. I've noticed SWAR implementation under checks for `<` and `>` characters while other vectored searches don't. SWAR is used as fallback to all SIMD implementations, so when the remaining buffer length is smaller than vector size (e.g `< 32`), caret chars won't be allowed. This might hurt consistency.
https://github.com/seanmonstar/httparse/blob/3308d1198dd8d817c4f0dcbc31f19a60a284a4d8/src/simd/swar.rs#L129-L156

nikneym:
This supersedes #202 and closes #201. SWAR validator now allows UTF8 characters as other SIMD backends. Tests are also updated.

### DIFF
--- a/src/simd/avx2.rs
+++ b/src/simd/avx2.rs
@@ -51,7 +51,7 @@ unsafe fn match_url_char_32_avx(buf: &[u8]) -> usize {
 
     // unsigned comparison dat >= LOW
     //
-    // We create a new via `_mm256_max_epu8` which compares vectors `dat` and `LOW`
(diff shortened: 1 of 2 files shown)


nikneym:
@seanmonstar can I receive a review? :eyes:

joelwurtz:
Thanks, i don't know how i miss that, doesn't the ci try to pass the test in all simd implementations ?

seanmonstar (approved):

PR #206, opened 2025-03-19
Update to use APIs up to MSRV

thaliaarchi:
Update to use APIs which are available now that the [MSRV is 1.47](https://github.com/seanmonstar/httparse/pull/192):

- `{integer}::from_ne_bytes` const since 1.44
- `{integer}::trailing_ones` stable since 1.46
- `str::strip_prefix` stable since 1.45

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -99,11 +99,11 @@ struct Version (u32, u32, u32);
 
 impl Version {
     fn parse(s: &str) -> Result<Version, String> {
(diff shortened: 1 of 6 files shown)


thaliaarchi:
I originally updated `Bytes::peek_n` to use const generics, but they were not stable in 1.47.

seanmonstar (approved):
Thanks so much!

PR #219, opened 2026-03-20
Remove code duplication in chunk size parsing

DemiMarie:
The logic for uppercase letters and lowercase letters can be merged.


seanmonstar (approved):
Ah yes, nice, thanks!

PR #220, opened 2026-06-21
refactor: use const generics from Rust 1.51

thaliaarchi:
Fixes a TODO now that the MSRV has been bumped past Rust 1.51.


seanmonstar (approved):
Nice!

PR #224, opened 2026-08-07
feat: add name_eq_ignore_case convenience method to Header

Aditya-9-6:
### Summary
Adds a zero-allocation `name_eq_ignore_case` convenience method to `Header<'a>` for case-insensitive header name comparisons against `&str` or `&[u8]`.

### Motivation
When inspecting parsed headers in web servers, proxies, and security middleware (e.g. `hyper`, `axum`, `pingora`), callers frequently need to check if a parsed `Header` matches a target header name case-insensitively (`"user-agent"`, `"authorization"`, `"x-forwarded-for"`). 

Previously, callers had to write `header.name.eq_ignore_ascii_case("target")` or convert byte slices manually. Adding `Header::name_eq_ignore_case` directly to `Header` provides clean API ergonomics and accepts any string or byte slice implementing `AsRef<[u8]>`.

### Key Changes
* **`Header::name_eq_ignore_case`**: Infallible method taking `&T` where `T: ?Sized + AsRef<[u8]>`. Performs length check fast-path followed by ASCII case-insensitive comparison.
* **Doctests & Unit Tests**: Added runnable doctests and unit tests in `src/lib.rs`.

### Verification
- `cargo test`: Passed all 54 unit tests cleanly.
- `cargo fmt`: Formatted cleanly.
- `Signed-off-by`: Yes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -747,6 +747,28 @@ impl fmt::Debug for Header<'_> {
     }
 }
 
+impl<'a> Header<'a> {
+    /// Case-insensitively compares this header's name against a target string or byte slice.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// let header = httparse::Header { name: "User-Agent", value: b"Mozilla/5.0" };
+    /// assert!(header.name_eq_ignore_case("user-agent"));
+    /// assert!(header.name_eq_ignore_case(b"USER-AGENT"));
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for looking to help! However, I don't want to add this new method. I think it's about the same amount for someone to use the method on `name`.

PR #225, opened 2026-08-12
Support all whitespace characters in header values

TimvdLippe:
This PR is part of a series of PRs across several crates, to eventually fix web-platform-test failures for Servo.

Servo currently fails [`/content-security-policy/generic/only-valid-whitespaces-are-allowed.html`](https://wpt.fyi/results/content-security-policy/generic/only-valid-whitespaces-are-allowed.html?product=servo) which is a test that checks whether various whitespace characters are properly parsed in the `Content-Security-Policy` header.

Currently, Servo times out on this test, since it attempts to send a HTTP request with Hyper. Hyper calls `httparse` to parse the response returned by the test server with, which includes the `Content-Security-Policy` with the special whitespace character. When `httparse` attempts to parse this, it fails since it contains characters it currently does not accept.

While RFC7230 (HTTP) does not allow such characters to be present, for web browsers [section 3.5](https://datatracker.ietf.org/doc/html/rfc7230#section-3.5) is relevant. That section states that for historical reasons (which is the case for web browsers), both `%x0B` and `%x0C` MAY be treated as regular space characters.

Since currently `httparse` rejects parsing, Servo is unable to implement that requirement. To fix this issue, the following changes are made:

1. In `httparse`, support parsing `%x0B` and `%x0C` in header values (https://github.com/seanmonstar/httparse/pull/225)
2. In `http`, consider `%x0B` and `%x0C` as opaque bytes for byte constructors (https://github.com/hyperium/http/pull/862)
3. In `hyper`, use `HeaderValue::from_bytes` rather than the generic `HeaderValue::from_maybe_shared_unchecked` to operate on the raw bytes (https://github.com/hyperium/hyper/pull/4155)
4. In `content-security-policy` update policy parsing to check if directive value is an ASCII string. It already handles `%x0B` and `%x0C` in `is_char_ascii_whitespace`, but with the HTTP changes it exposed this other missing check (https://github.com/rust-ammonia/rust-content-security-policy/pull/74)
5. In Servo, replace `header.to_str()` with `str::from_utf8(header.as_bytes())` (https://github.com/servo/servo/pull/47194)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,7 +88,7 @@ pub(crate) fn is_header_name_token(b: u8) -> bool {
 
 
 static HEADER_VALUE_MAP: [bool; 256] = byte_map!(
-    b'\t' | b' '..=0x7e | 0x80..=0xFF
+    b'\t' | b' '..=0x7e | 0x80..=0xFF | 0x0B..=0x0C
 );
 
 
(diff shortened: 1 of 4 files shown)


TimvdLippe:
@seanmonstar Friendly ping, do you mind looking at this group of PRs? All in all, this fixes a whole bunch of test failures for Servo and I would love to continue making Servo more web-compatible.

seanmonstar:
Hey! Interesting history, indeed.

I'll admit, I'm reluctant to move forward with this. It seems to weaken the parser for everyone, and the changed in the other libraries look like potential reduction in performance and guarantees of `HeaderValue`.

For the gain of what I assume is a very rare feature. Is simply marking that test as wont-fix an option?

TimvdLippe:
This affects at least 18 tests across various web platform features in Servo. You can see the full list under the "Stable unexpected results" in https://github.com/servo/servo/pull/47194#issuecomment-5265609427

I understand your reluctance. Would you be more comfortable if we put this behind a feature flag instead?

TimvdLippe:
When discussing what the value of the change is, @jdm pointed out this breaks real websites: https://github.com/webcompat/web-bugs/issues/18902 Therefore, we consider this at Servo an important quirk to fix, such that those websites load correctly.

seanmonstar:
OK, I took another look and read over the various specs and tests. 🤓 

I notice that the spec (RFC 9110 is the latest, but 7230 said so as well) indicates that while we _might_ allow this, all such bytes should be interpreted as SP. So, if that is acceptable, I'd say the code in hyper should do similar to the `obs_fold` treatment: if enabled, then look for these extra bytes, and convert them to SP. Then nothing else needs to worry about odd bytes appearing where they weren't supposed to.

It's not really acceptable to have a strong type over a string, like `HeaderValue`, promise it will only hold certain bytes, and then in some situations they might appear. Other code could have expected otherwise.

But, to my ~~delight~~, it looks like the WPT case actually expects the VT to survive all the way through, and **not** be treated as SP. Which would make what I suggested not work. Am I understanding correctly? 😭

TimvdLippe:
Well, to some extent I am delighted to realise that we went down the same rabbit hole. So yes, I started at Servo and seeing the requests to fail, to then via Hyper drill down further until I realised the whole byte needs to survive from parsing all the way to Servo to make the logic work. Unfortunately, that means the space character replacement also wouldn't work.

CSP is directly contradicting HTTP as you say and I want to file a spec issue for that. Still, I think browsers are unlikely to change behavior for a VT, given that they have been working this way ever since and there is actual web impact.

I am not thrilled by it either, but at least this is a good finding to understand what's missing in CSP and what browsers actually do.

